PR #4, opened 2017-01-25
pipeline changes/fixes

MartinNowak:
- run parallel around node (to throttle and distribute parallelism across executors)
- directly load ProjectsJenkinsfile
- more project testing
- couple of fixes

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -111,25 +111,19 @@ node { // for now whole pipeline runs on one node because no slaves are present
     }
 
     stage("Package distribution") {
-        // ideally this step should be in sync with release packaging scripts
-        // but that requires a lot of additional work to set up virtualization
-        // for different platforms and separate manual bits from automatic in
-        // actual packaging script. Thus for now most simple tarball is
-        // generated - just enough to run downstream project tests
+        // ideally this step should be in sync with the release tars
+        sh '''#!/usr/bin/env bash
+            set -ueo pipefail
 
-        sh "mkdir -p distribution/{bin,imports,libs}"
-
-        sh "cp dmd/src/dmd distribution/bin/"
-        writeFile file: 'distribution/bin/dmd.conf', text: '''[Environment]
-    DFLAGS=-I%@P%/../imports -L-L%@P%/../libs -L--export-dynamic -L--export-dynamic -fPIC'''
-        sh "cp dub/bin/dub distribution/bin/"
-        sh "cp tools/generated/linux/64/rdmd distribution/bin/"
-
-        sh "cp -r phobos/{etc,std} distribution/imports/"
-        sh "cp -r druntime/import/* distribution/imports/"
-        sh "cp phobos/generated/linux/release/64/libphobos2.a distribution/libs"
-
-        sh "tar -cf distribution.tar distribution"
+            rm -rf distribution
+            mkdir -p distribution/{bin,imports,libs}
+            cp --recursive --link dmd/src/dmd dub/bin/dub tools/generated/linux/64/rdmd distribution/bin/
+            cp --recursive --link phobos/etc phobos/std druntime/import/* distribution/imports/
+            cp --recursive --link phobos/generated/linux/release/64/libphobos2.a distribution/libs/
+            echo '[Environment]
+DFLAGS=-I%@P%/../imports -L-L%@P%/../libs -L--export-dynamic -L--export-dynamic -fPIC' > distribution/bin/dmd.conf
+            tar -cf distribution.tar distribution
+        '''
         archiveArtifacts artifacts: 'distribution.tar', onlyIfSuccessful: true
     }
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,8 +12,9 @@
     `git clean -fdx`) in the process.
  **/
 def cleanCheckout (repo_url, git_ref = "master") {
-    git poll: false, branch: "master",
-        extensions: [[$class: 'CleanBeforeCheckout']], url: repo_url
+    checkout poll: false,
+        scm: [$class: 'GitSCM', branches: [[name: git_ref]], doGenerateSubmoduleConfigurations: false,
+              extensions: [[$class: 'CleanBeforeCheckout']], submoduleCfg: [], userRemoteConfigs: [[url: repo_url]]]
 }
 
 /**
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,6 +80,7 @@ node { // for now whole pipeline runs on one node because no slaves are present
         def action = { sh "make -f posix.mak RELEASE=1 AUTO_BOOTSTRAP=1" }
 
         dir('dmd',      action)
+        dir('dmd/src', { sh 'make -f posix.mak AUTO_BOOTSTRAP=1 dmd.conf' })
         dir('druntime', action)
         dir('phobos',   action)
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -122,9 +122,8 @@ node { // for now whole pipeline runs on one node because no slaves are present
             cp --recursive --link phobos/generated/linux/release/64/libphobos2.a distribution/libs/
             echo '[Environment]
 DFLAGS=-I%@P%/../imports -L-L%@P%/../libs -L--export-dynamic -L--export-dynamic -fPIC' > distribution/bin/dmd.conf
-            tar -cf distribution.tar distribution
         '''
-        archiveArtifacts artifacts: 'distribution.tar', onlyIfSuccessful: true
+        stash name: "dlang-build", includes: "distribution/**"
     }
 
 }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -80,6 +80,7 @@ node { // for now whole pipeline runs on one node because no slaves are present
         def action = { sh "make -f posix.mak RELEASE=1 AUTO_BOOTSTRAP=1" }
 
         dir('dmd',      action)
+        dir('dmd/src', { sh 'make -f posix.mak AUTO_BOOTSTRAP=1 dmd.conf' })
```
Dicebot: Didn't know of this makefile target :+1:

Jenkinsfile
```diff
@@ -126,8 +126,8 @@ DFLAGS=-I%@P%/../imports -L-L%@P%/../libs -L--export-dynamic -L--export-dynamic
...
         archiveArtifacts artifacts: 'distribution.tar', onlyIfSuccessful: true
     }
-}
 
-stage("Test downstream projects") {
-    build job: 'dlangci-downstream'
 }
+
+// executed from workspace root
+load('dlangci/ProjectsJenkinsfile')
```
Dicebot: But workspace root is supposed to be `dlangci` itself by default, how does it come to be in a separate folder?

ProjectsJenkinsfile
```diff
@@ -90,26 +90,13 @@ def dub_projects = [
...
-    }
+    unstash name: "dlang-build"
 
-    stage('Test') {
+    stage('Test Projects') {
         parallel mapSteps(dub_projects, { name ->
+            clone(name)
             withEnv(["PATH=${env.WORKSPACE}/distribution/bin:${env.PATH}"]) {
-                sh "dub build"
+                sh "dub test"
```
Dicebot: Please do both `dub test` and `dub build`, this is quite important in general case.

ProjectsJenkinsfile
```diff
@@ -90,26 +90,13 @@ def dub_projects = [
...
-                sh "dub build"
-            }
-        })
-    }
+    unstash name: "dlang-build"
 
-    stage('Test') {
+    stage('Test Projects') {
         parallel mapSteps(dub_projects, { name ->
+            clone(name)
```
Dicebot: `cleanCheckout "https://github.com/${name}.git"`

`clone` function was only used as `mapSteps` predicate and can be removed in absence of dedicated `clone` step

Jenkinsfile
```diff
@@ -12,8 +12,9 @@
     `git clean -fdx`) in the process.
  **/
 def cleanCheckout (repo_url, git_ref = "master") {
-    git poll: false, branch: "master",
-        extensions: [[$class: 'CleanBeforeCheckout']], url: repo_url
+    checkout poll: false,
+        scm: [$class: 'GitSCM', branches: [[name: git_ref]], doGenerateSubmoduleConfigurations: false,
+              extensions: [[$class: 'CleanBeforeCheckout']], submoduleCfg: [], userRemoteConfigs: [[url: repo_url]]]
```
Dicebot: Can you comment on the issue here? I had impression that `git` is simply a wrapper on top of `checkout` with bunch of preconfigured settings.

Jenkinsfile
```diff
@@ -54,9 +54,9 @@ def clone (name) {
 
 def test (name) {
     if (name == 'dmd')
-        sh "make -j 4 -f posix.mak test MODEL=64"
+        sh "make -f posix.mak test MODEL=64 --jobs=4"
```
Dicebot: This parallelization is only truly relevant for dmd tests, I am not sure if this is a good idea to enable it for everything because it will clash with plain build runner parallelization, resulting in extra scheduling contention OS-wise. How many cores does the jenkins server have?

ProjectsJenkinsfile
```diff
@@ -89,16 +89,19 @@ def dub_projects = [
...
 ]
 
-node { // for now whole pipeline runs on one node because no slaves are present
-
-    unstash name: "dlang-build"
-
-    stage('Test Projects') {
-        parallel mapSteps(dub_projects, { name ->
-            clone(name)
+def runTest(name) {
```
Dicebot: Please move `mapSteps` predicate to the comment section reserved for them

ProjectsJenkinsfile
```diff
@@ -84,19 +84,48 @@ def dub_projects = [
...
+   "rejectedsoftware/vibe.d#libevent",
+   "rejectedsoftware/vibe.d#libasync",
    "repeatedly/mustache-d",
    "s-ludwig/taggedalgebraic",
 ]
 
 def runTest(name) {
     def n = name // to fix issues with closure
+    def repo = name.split('#')[0]
+    echo repo
```
Dicebot: That means `rejectedsoftware/vibe.d#libevent` and `rejectedsoftware/vibe.d#libasync` will be cloned and tested in same workspace folder, doubt this is what you intended. I'd prefer to sacrifice parallelism here and keep one `rejectedsoftware/vibe.d` project which has two `./travis-ci.sh` calls.

ProjectsJenkinsfile
```diff
@@ -15,7 +15,7 @@ def cleanCheckout (repo_url, git_ref = "master") {
               extensions: [[$class: 'CleanBeforeCheckout']], submoduleCfg: [], userRemoteConfigs: [[url: repo_url]]]
 
     def LATEST = sh (
-        script: 'git tag -l | egrep "^v[0-9]+\\.[0-9]+\\.[0-9]+.*$" | sort --version-sort | tail -n 1',
+        script: 'git tag -l | egrep "^v?[0-9]+\\.[0-9]+\\.[0-9]+$" | sort --version-sort | tail -n 1',
```
Dicebot: Not good. You don't only ignore pre-prelease tags, but also any release with metadata (http://semver.org/#spec-item-10), make it `(\+.+)?` at least

Jenkinsfile
MartinNowak: [Dedibox® LT 2016 - Online.net](https://www.online.net/en/dedicated-server/dedibox-lt)
1x Intel® Xeon® E3 1245 v5 
4 C / 8T @3.5 Ghz cache L3 8MB, x64, VT

Parallelizing those make sense to reduce the latency of test results. I don't think we have that many PRs to test at the same time. And I also guess that it's more efficient to run druntime->phobos->dmd tests one after another rather than in parallel, can check that.

Jenkinsfile
MartinNowak: The git command simply doesn't process take/process an extensions parameter.
https://jenkins.io/doc/pipeline/steps/git/#git-git

Jenkinsfile
MartinNowak: I've changed it b/c it was a mess to have a loose git folder as root and lots of other projects inside of that.
I'd rather if we put every clone under `ghrepos/` with their github path, e.g. `ghrepos/Dicebot/dlangci`.
That way caching clones will be less confusing if multiple jobs/stages interpose their repo layout.

ProjectsJenkinsfile
MartinNowak: What's supposed to be tested by that? Building without `-unittest`? Wouldn't that always be a subset?
Remember we're not testing dub projects here, but dmd/druntime/phobos.

ProjectsJenkinsfile
MartinNowak: No it's separate checkouts.
Vibe.d is the longest running test in the suite, but libasync contributes only a small part.
Let's remove that and maybe parallize vibe.d's test suite at some point instead.

Jenkinsfile
Dicebot: > I've changed it b/c

I have probably missed the line that changed it - Jenkins will always checkout repo that triggers the build by default with organization plugin and I was looking for a way to disable that. How do you do it?

ProjectsJenkinsfile
Dicebot: `dub test` runs `unittest` configuration if present (http://code.dlang.org/package-format?lang=sdl):

> The configuration name "unittest" has a special meaning - if a configuration with this name is present, it will be used by default when executing dub test. It can be assumed to have the -unittest flag present (through the "unittest" build type). 

That can be defined to use a different sources and/or compiler flags. Plain `dub build`, on the other hand, ensures that primary package target compiles with its default flags.

ProjectsJenkinsfile
Dicebot: > No it's separate checkouts.

How does it work? They are part of same workspace and have same clone path - I'd expect those checkouts to have a race upon parallelization.

Jenkinsfile
Dicebot: >_<

Jenkinsfile
Dicebot: I see, let's measure it on final setup https://github.com/Dicebot/dlangci/issues/8

Jenkinsfile
MartinNowak: So far I've only manually build triggered builds, but the git/SCM plugin has a way to do that.
http://stackoverflow.com/a/9791873

PR #12, opened 2017-02-04
Support non-master PR base branch

Dicebot:
Example: https://ci2.dawg.eu/blue/organizations/jenkins/dlang-test-organization%2Fdmd/detail/PR-1/32/pipeline/28

Not very elegant but I don't think base branch info can be retrieved via Jenkins own metadata

### DIFF
--- a/pipeline.groovy
+++ b/pipeline.groovy
@@ -23,7 +23,8 @@ def clone (repo_url, git_ref = "master") {
 
 /**
     Function to checkout upstream that has triggerred current
-    pipeline, for example PR branch.
+    pipeline, for example PR branch. For PRs this will already merge
+    with the base branch.
 
     Requires administrator approval for allow access to:
 
@@ -97,7 +98,7 @@ def mapSteps (names, action) {
 *******************************************************************************/
 
 def getSources (name) {
-    def git_ref = 'master'
+    def base_branch = 'master'
     def pr_repo
     def pr_id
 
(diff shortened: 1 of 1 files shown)


pipeline.groovy
```diff
@@ -97,19 +97,35 @@ def mapSteps (names, action) {
...
-    def pr_repo = match ? match[0][1] : ""
+    if (match) {
+        pr_repo = match[0][1]
+        pr_id = match[0][2]
+    }
     match = null
 
+    if (pr_repo && pr_id) {
+        def api_url = "https://api.github.com/repos/dlang/$pr_repo/pulls/$pr_id"
+        def extract_cmd = /python3 -c "import sys, json; print(json.load(sys.stdin)['base']['ref'])"/
```
MartinNowak: Can use [jq](https://stedolan.github.io/jq/) for such stuff, already preinstalled.
`jq -r '.base.ref'`
Also see https://github.com/dlang/dmd/blob/4f1425c1c7b07ffd72997151c614d51f63ddc082/circleci.sh#L55.

Dicebot: Yep, I was not sure if it is installed and went with python as more widespread option. Will adjust.

MartinNowak:
Looks fairly good, cloneUpstream seems to be missing a merge with the base branch.
```
[[$class: 'PreBuildMerge', options: [fastForwardMode: 'FF', mergeRemote: 'git://github.com/dlang/dmd.git', mergeStrategy: 'MergeCommand.Strategy', mergeTarget: base_branch]],
```
Also see https://github.com/dlang/dmd/blob/4f1425c1c7b07ffd72997151c614d51f63ddc082/circleci.sh#L54-L79.

MartinNowak:
Ah, already seems to work, https://ci2.dawg.eu/blue/organizations/jenkins/dlang-test-organization%2Fdmd/detail/PR-2/1/pipeline/25#step-45-log-26.

PR #14, opened 2017-02-12
fix #13 - reuse same workspace for dmd, druntime, ... jobs

### LINKED ISSUE #13: Github Organization/Branch Sources plugin creates unique workspaces per job (PR/branch)
This goes somewhat against our idea to cache the checkouts.
It seems the impedance mismatch originates from us wanting to run the same testsuite for multiple repositories.

MartinNowak:


MartinNowak:
See #13.

pipeline.groovy
```diff
@@ -179,6 +179,12 @@ def testDownstreamProject (name) {
 *******************************************************************************/
 
 node { // for now whole pipeline runs on one node because no slaves are present
+    /* Use the same workspace, no matter what job (dmd, druntime,...)  triggered
+     * the build.  The workspace step will take care of concurrent test-runs and
+     * allocate additional workspaces if necessary.  This setup avoids to
+     * reclone repos for each test-run.
+     */
+    ws: 'dlangci'
```
Dicebot: Interesting that documentation for `ws` seems to imply that initial workspace will be allocated anyway from `node` - the custom one will be switched to after. How does it work in practice?

MartinNowak: It's based on the job name (see https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/8f7dac968c97e2a123befc2419f16e24234ca785/src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java#L45-L58).
`ls /var/lib/jenkins/workspace` @ runner-01.ci2.lxd
```
-test-organization_dmd_PR-2-43RSTKUS5XTSIZNXNWDK5S56E547FOJ4WDFVZTW5CQE7DI65LTJQ
-test-organization_dmd_PR-2-43RSTKUS5XTSIZNXNWDK5S56E547FOJ4WDFVZTW5CQE7DI65LTJQ@2
-test-organization_dmd_PR-2-43RSTKUS5XTSIZNXNWDK5S56E547FOJ4WDFVZTW5CQE7DI65LTJQ@2@tmp
-test-organization_dmd_PR-2-43RSTKUS5XTSIZNXNWDK5S56E547FOJ4WDFVZTW5CQE7DI65LTJQ@3
-test-organization_dmd_PR-2-43RSTKUS5XTSIZNXNWDK5S56E547FOJ4WDFVZTW5CQE7DI65LTJQ@4
-test-organization_dmd_PR-3-GOQOXUB3V5PYYBYOIE2XZ3DSWD4IWUVUIOVANGFNIK7FT3OXDKRQ
-test-organization_dmd_PR-3-GOQOXUB3V5PYYBYOIE2XZ3DSWD4IWUVUIOVANGFNIK7FT3OXDKRQ@tmp
```

MartinNowak: So this step will just create and switch to `/var/lib/jenkins/workspace/dlang`, leaving behind an empty job-specific workspace.

Dicebot: Are you sure about "empty" bit? It has to do initial clone to get `Jenkinsfile` for pipeline to run.

Dicebot: Ok, checked the code, I see now - it only affects node workspaces while main master checkout is processed separately.

MartinNowak:
Weird default to cherry-pick instead of merge ;), see 22e0b44e7961e94b428b64c44c17cbd1a4f347ec.

PR #15, opened 2017-02-15
some improvements on project tests

MartinNowak:
- similar env to install.sh/Travis-CI
- run .travis.yml script
- put shared libs into distribution
- clean dub build caches
- .dub folder per workspace

### DIFF
--- a/pipeline.groovy
+++ b/pipeline.groovy
@@ -140,7 +140,9 @@ def testDownstreamProject (name) {
                     "LIBRARY_PATH+LIB=${env.WORKSPACE}/distribution/libs",
                     "LD_LIBRARY_PATH+LIB=${env.WORKSPACE}/distribution/libs",
                     'DC=dmd',
-                    'DMD=dmd'
+                    'DMD=dmd',
+                    // set HOME to separate concurrent ~/.dub user paths
+                    "HOME=${env.WORKSPACE}"
                 ]) {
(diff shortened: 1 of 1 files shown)


pipeline.groovy
```diff
@@ -1,3 +1,5 @@
+import groovy.json.JsonSlurper
```
Dicebot: Unused leftover?

pipeline.groovy
```diff
@@ -279,4 +279,8 @@ DFLAGS=-I%@P%/../imports -L-L%@P%/../libs -L--export-dynamic -L--export-dynamic
     stage ('Test Projects') {
         parallel mapSteps(dub_projects, this.&testDownstreamProject)
     }
+
+    stage ('Cleanup') {
+        sh 'find $HOME/.dub/packages -type d -name .dub -exec rm -r {} +'
```
Dicebot: Isn't this now part of dub core functionality (to rebuild projects when compiler version changes)?

MartinNowak: It doesn't yet use the `dmd --version` reported version (which includes a commit hash).
https://github.com/dlang/dub/blob/038fc1b471656ed76217f7f9da81a361a6f68b10/source/dub/generators/build.d#L344

MartinNowak: Even then we'd want to regularly clean this.

PR #21, opened 2017-02-25
make project tests pass

MartinNowak:


MartinNowak:
Same workarounds as on ci.dawg.eu, finally testing everything and passing.
https://ci2.dawg.eu/blue/organizations/jenkins/dlang/detail/dlang/150/pipeline

PR #28, opened 2017-03-03
use fixed vibe-d for diet-ng tests

MartinNowak:


pipeline.groovy
```diff
@@ -130,6 +130,16 @@ def getSources (name) {
     }
 }
 
+def test_travis_yaml () {
+    def script = 'dub test --compiler=$DC'
+    if (fileExists('.travis.yml')) {
+        def travis_script = sh(script: 'get_travis_test_script', returnStdout: true).trim()
```
wilzbach: I couldn't find any reference to `get_travis_test_script`, so I assume it's a magic binary/script that's within the `PATH`?

pipeline.groovy
```diff
@@ -191,17 +201,16 @@ def testDownstreamProject (name) {
                 case 'BlackEdder/ggplotd':
                     // workaround https://github.com/BlackEdder/ggplotd/issues/34
                     sh 'sed -i \'s|auto seed = unpredictableSeed|auto seed = 54321|\' source/ggplotd/example.d'
-                    sh 'dub test'
+                    test_travis_yaml()
+                    break;
+
+                case 'rejectedsoftware/diet-ng':
+                    sh 'sed -i \'s|dependency "vibe-d".*|dependency "vibe-d" version="0.7.31-beta.2"|\' examples/htmlserver/dub.sdl'
```
wilzbach: This might break again when Sönke adds a new test/example - how about  `**/*.sdl` ?

pipeline.groovy
MartinNowak: https://github.com/MartinNowak/dlangci/commit/fe388f742a6fc44c7aa8ec0de76153673f6bd211
The [server](https://github.com/MartinNowak/dlangci/commits/server) is going to be merged when the remaining 2 topics are done.

pipeline.groovy
MartinNowak: Let's rather be explicit.

PR #43, opened 2017-06-13
Add CyberShadow/ae

CyberShadow:
This library is used for Digger, DAutoTest and DFeed.

### DIFF
--- a/pipeline.groovy
+++ b/pipeline.groovy
@@ -303,6 +303,7 @@ DFLAGS=-I%@P%/../imports -L-L%@P%/../libs -L--export-dynamic -L--export-dynamic 
     def dub_projects = [
         "Abscissa/libInputVisitor",
         "BlackEdder/ggplotd",
+        "CyberShadow/ae",
         "DerelictOrg/DerelictFT",
         "DerelictOrg/DerelictGL3",
         "DerelictOrg/DerelictGLFW3",


CyberShadow:
Changes to the CI itself aren't tested?

wilzbach:
> Changes to the CI itself aren't tested?

Nope not that I would know ...

MartinNowak:
> Nope not that I would know ...

Changes against the CI repo itself would need to be integrated into https://github.com/dlang/phobos/blob/master/Jenkinsfile.
Alternatively we could use a bootstrap groovy file that loads the real pipeline script, see #44.

MartinNowak:
Or just use a different Jenkinsfile in the ci repo :).

MartinNowak:
Please check the output of https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fci/detail/master/2/pipeline/162#step-316-log-1 whether everything works as expected.
We're simply reading the script from the `.travis.yml` by default (with `dub test --compiler=$DC` as fallback just as on Travis-CI).

The workers actually do run on Btrfs btw ;).

CyberShadow:
Yep, looks great! 

Perhaps I should exclude the network tests to avoid random failures due to changes in network conditions. Any advice on how do detect if it's running on the Jenkins CI from the D unittest?

> The workers actually does run on Btrfs btw ;).

It reads /proc/self/mounts and looks for the directory `pwd` is in, if that's somehow wrong/unavailable then it won't be detected.

PR #44, opened 2017-06-15
Test updates to CI repo itself

MartinNowak:
Maybe using some intermediate script that loads the real pipeline to keep this out of the dlang repos.
https://github.com/jenkinsci/workflow-remote-loader-plugin

Just use the Jenkinsfile, repo should already get cloned by Jenkins.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,11 +1,21 @@
 #!/bin/env groovy
 
-node {
-    stage ("Trigger") {
-        dir ("dlang/ci") {
-            git "https://github.com/dlang/ci.git"
-        }
-
-        load "dlang/ci/pipeline.groovy"
-    }
+def cloneUpstream () {
+    checkout(scm: [
+        $class: 'GitSCM',
+        branches: scm.branches,
+        extensions: scm.extensions + [[$class: 'CleanBeforeCheckout']],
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,11 +1,21 @@
...
+    ])
+}
 
-        load "dlang/ci/pipeline.groovy"
+def pipeline
+node {
+    dir('dlang/ci') {
+        cloneUpstream()
     }
+    pipeline = load 'dlang/ci/pipeline.groovy'
```
MartinNowak: This might circumvent the trusted Jenkinsfile source policy of Jenkins. Which is fine since the workers are encapsulated and there isn't much to protect on that server either.

MartinNowak: Indeed it does, see https://ci.dlang.io/job/dlang-org/job/ci/job/PR-45/1/console, first takes Jenkinsfile from master, then pipeline.groovy from the PR commit.
This is good, as it actually tests the correct pipeline script of the PR.

CyberShadow:
Nice. A possible improvement would be to test specific tags of every tested repo, so that it becomes easy to distinguish breakages due to updates from the D side or the project's side. Periodically the tags could be updated if the new tags pass the tests.

PR #48, opened 2017-06-26
provision dub-registry mirror on code-mirror.dlang.io

MartinNowak:
Should reduce failures due to network connectivity issues.

### DIFF
--- a/ansible/ci.yml
+++ b/ansible/ci.yml
@@ -4,6 +4,12 @@
     - common
     - lxd_host
 
+- hosts: dub-registry.ci.lxd
+  vars:
+    - dub_registry_http_port: 8080
+  roles:
+    - dub_registry_mirror
+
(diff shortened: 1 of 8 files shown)


MartinNowak:
FYI @wilzbach

wilzbach:
> FYI @wilzbach

Yep seen it, but was/am still on the go. This looks pretty cool. Thanks!
The next logical step would be to add the mirror to DUB as a list of fallback mirrors, so that Travis builds etc. also stop failing randomly.
Are you okay with it being publicly used as a fallback?
I will also look into running a mirror myself and maybe even adding a ""deploy on heroku" button, so that it's dead simple for people to help out and run a mirror.

MartinNowak:
Let's see for a while how that goes before hard-coding it into clients.
Also we prolly shouldn't jump to the other extreme, a huge number of decentralized caches, right away.

PR #49, opened 2017-06-29
use dub clean --all-packages instead of manual delete

MartinNowak:


ZombineDev:
> /var/lib/jenkins/dlang_projects@tmp/durable-bd130c46/script.sh: dub: not found

Hmm, looks like you have more work to do

PR #72, opened 2017-09-25
use concurrent G1GC

MartinNowak:
- hopefully improving horrible load times a bit


MartinNowak:
Definitely a difference between night and day, still slow somewhat.

I don't want to purge old logs as that will invalidate test result links.
Let's see whether there is some better archive solution.

BTW, I've seen a couple of ConcurrentModificationException recently hanging a build, not sure whether those came with some update or with the new on-demand slave. Hope it's not going to be too disturbing.

PR #78, opened 2017-10-06
try to enable address and undefined sanitizer for dmd

MartinNowak:


MartinNowak:
:( linker order on Ubuntu matters and dmd doesn't even preserve it.
[Issue 15574 – wrong order of linker arguments breaks with as-needed](https://issues.dlang.org/show_bug.cgi?id=15574)

MartinNowak:
The overall memory overhead and the very few dmd issues make this not so interesting.
Also a weird idea to only run this kind of test on dmd here, but not on the dmd's test-suite (auto-tester).

PR #80, opened 2017-10-06
use Library to load pipeline.groovy script

MartinNowak:
- avoids checkout of the whole ci repo
- temporarily added symlink for now to not break other dlang Jenkinsfiles


MartinNowak:
Remains a joy to use Jenkins :/.
I'll update all the other repo Jenkinsfiles in lockstep, once this has been tested.

PR #84, opened 2017-10-14
Add 'dlang/undeaD' to the pipeline

ZombineDev:
Removing functions from Phobos can easily break undeaD. We need to prevent that if we want to continue to able to aggressively deprecate functionality that's not up to standards of Phobos.

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -329,6 +329,7 @@ DFLAGS=-I%@P%/../imports -L-L%@P%/../libs -L--export-dynamic -L--export-dynamic 
         "DerelictOrg/DerelictGL3",
         "DerelictOrg/DerelictGLFW3",
         "DerelictOrg/DerelictSDL2",
+        "dlang/undeaD",
         "DlangScience/scid",
         "Netflix/vectorflow",
         "ariovistus/pyd",


ZombineDev:
@MartinNowak Let me know if there's anything else I need to do as I am very far from being experienced Jenkins developer.

jmdavis:
I would point out that maintaining undead goes against the whole purpose of deprecating stuff. We don't want to have to maintain this stuff. It _will_ break upon occasion, and that's life. The purpose of undead is to try and make it easier for folks to upgrade their projects if they waited stupidly long to upgrade, and the functionality they're using is gone from Phobos, but if we're having to avoid breaking undead, then that's a serious problem. Anyone who really wants to use this functionality long term should be copying it into their own projects, not using undead long term.

ZombineDev:
> I would point out that maintaining undead goes against the whole purpose of deprecating stuff. 

On the contrary! I am all for deprecating functions that are not up to the standards of Phobos. In the long term I think we all want to have a clean and beautiful standard library, and I see the deprecation process as the main tool for correcting design mistakes.

> We don't want to have to maintain this stuff.

We can't allow every new release to break the ecosystem anymore. That's just not practical nor economical. Maintenance is just not an issue for us. Fixing a problem with undead takes the average Phobos contributor less than 30 minutes. Figuring out why a dub package X, that uses dub package Y, that uses dub package Z, that uses undead which breaks with DMD 2.077.0 may take the average D user up to several days to figure out after complaining on the forums and someone finally deciding to help him. And if I were a new user and experience this on my first day of evaluating the ecosystem, I may as well quit.

> The purpose of undead is to try and make it easier for folks to upgrade their projects if they waited stupidly long to upgrade, and the functionality they're using is gone from Phobos

No need to repeat this point every time, I doubt you'll find anyone on the team who doesn't know that.

> but if we're having to avoid breaking undead, then that's a serious problem

Where's the problem? As I said, it's not fair to put the burden on figuring out why something in a long chain of dependencies doesn't work on the average user. If you do a simple calculation, time lost by everyone losing days to debug why something doesn't work outweighs many times the minutes "lost" in moving removed functionality to undead by a single contributor.

> Anyone who really wants to use this functionality long term should be copying it into their own projects, not using undead long term.

Look, I have no intention in using undead in any of my personal projects, however I nor most anyone here has the time to fix every single third-party package out there.

ZombineDev:
Ping @MartinNowak are you ok moving forward with this?

MartinNowak:
The main requirement for any ci.dlang.org project is that it is reasonably maintained, as we cannot take over maintenance for those projects. Looks like some people do maintain it, https://github.com/dlang/undeaD/commits/master, so it qualifies.
We can always remove the project if it becomes to much of a burden.

ZombineDev:
Thanks!

PR #88, opened 2017-11-26
Start testing Sociomantic ocean in pipeline

mihails-strasuns:
Recently (https://github.com/sociomantic-tsunami/ocean/pull/321) we have enabled CI testing with vanilla upstream DMD compiler for WIP ocean major branch, as 2.077 provided necessary `Throwable.message()`. 

Would appreciate getting it to the list of automatically tested projects ;)

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -151,7 +151,7 @@ def testDownstreamProject (name) {
                 if (repo == 'rejectedsoftware/vibe.d') {
                     clone("https://github.com/${repo}.git", 'v0.8.1-rc.1')
                 } else if (repo == "sociomantic-tsunami/ocean") {
-                    clone("https://github.com/${repo}.git", 'v4.0.0-alpha.1')
+                    clone("https://github.com/${repo}.git", 'v4.0.0-alpha.2')
                 } else {
                     cloneLatestTag("https://github.com/${repo}.git")
                 }
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -151,7 +151,7 @@ def testDownstreamProject (name) {
                 if (repo == 'rejectedsoftware/vibe.d') {
                     clone("https://github.com/${repo}.git", 'v0.8.1-rc.1')
                 } else if (repo == "sociomantic-tsunami/ocean") {
-                    clone("https://github.com/${repo}.git", "v4.0.0-alpha.0")
+                    clone("https://github.com/${repo}.git", 'v4.0.0-alpha.0')
                 } else {
                     cloneLatestTag("https://github.com/${repo}.git")
                 }
@@ -210,11 +210,11 @@ def testDownstreamProject (name) {
                     break;
 
                 case 'sociomantic-tsunami/ocean':
-                    sh """
+                    sh '''
                     git submodule update --init
                     make d2conv V=1
                     make test V=1 DVER=2 F=production
-                    """
+                    '''
                     break;
 
                 default:
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -151,7 +151,7 @@ def testDownstreamProject (name) {
                 if (repo == 'rejectedsoftware/vibe.d') {
                     clone("https://github.com/${repo}.git", 'v0.8.1-rc.1')
                 } else if (repo == "sociomantic-tsunami/ocean") {
-                    clone("https://github.com/${repo}.git", 'v4.0.0-alpha.0')
+                    clone("https://github.com/${repo}.git", 'v4.0.0-alpha.1')
                 } else {
                     cloneLatestTag("https://github.com/${repo}.git")
                 }
@@ -211,10 +211,9 @@ def testDownstreamProject (name) {
 
                 case 'sociomantic-tsunami/ocean':
                     sh '''
-                    sed -i '/^override DFLAGS += -de$/d' Build.mak
                     git submodule update --init
                     make d2conv V=1
-                    make test V=1 DVER=2 F=production
+                    make test V=1 DVER=2 F=production ALLOW_DEPRECATIONS=1
                     '''
                     break;
 
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -211,14 +211,9 @@ def testDownstreamProject (name) {
 
                 case 'sociomantic-tsunami/ocean':
                     sh """
-                    wget -O libebtree.deb https://bintray.com/sociomantic-tsunami/dlang/download_file?file_path=libebtree6_6.0.socio6-xenial_amd64.deb
-                    dpkg -x libebtree.deb libebtree/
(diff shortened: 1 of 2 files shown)


mihails-strasuns:
`./src/ocean/text/util/StringSearch.d(265): Deprecation: cannot subtract pointers to different types: `void*` and `const(char)*` used to work with 2.077 :(

mihails-strasuns:
Good, now fails because ocean is built by default with deprecations-as-errors, which, of course, is not suitable for regression testing with upstream. Will require small tweak in our makefile.

MartinNowak:
Nice, will have a look at pre-installing the dependencies with ansible, assuming that they don't change too often.

mihails-strasuns:
> Nice, will have a look at pre-installing the dependencies with ansible, assuming that they don't change too often.

Yes, I think at this point they change less often than once a year. Alternatively, you can enable docker on the CI server and script can fetch the very same image we use for our own CI tests.

mihails-strasuns:
No problem, whatever suits you best. Btw there is no need to add workarounds for deprecation bit - I will update the PR as soon as https://github.com/sociomantic-tsunami/ocean/pull/374 is merged which should be pretty soon :)

vars/runPipeline.groovy
```diff
@@ -206,6 +209,15 @@ def testDownstreamProject (name) {
                     sh 'dub test --compiler=$DC'
                     break;
 
+                case 'sociomantic-tsunami/ocean':
+                    sh '''
+                    sed -i '/^override DFLAGS += -de$/d' Build.mak
+                    git submodule update --init
+                    make d2conv V=1
+                    make test V=1 DVER=2 F=production LDFLAGS=-L./libebtree/usr/lib/
```
mihails-strasuns: Shouldn't need `LDFLAGS` here if dependencies are pre-installed

mihails-strasuns:
Tagged https://github.com/sociomantic-tsunami/ocean/releases/tag/v4.0.0-alpha.1 and pushed commit to switch to it

MartinNowak:
Thx, I'd prefer help with the failing tests.
https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fci/detail/PR-88/8/pipeline/188#step-342-log-368
```
ocean.core.Test.TestException@/var/lib/jenkins/dlang_projects@2/sociomantic-tsunami/ocean/test/sysstats/main.d(68): expression 'nan >= 0' evaluates to false
----------------
./src/ocean/core/Enforce.d:416 void ocean.core.Enforce.enforceImpl!(">=", float, int).enforceImpl(lazy Exception, float, int, immutable(char)[], int) [0x6281dd]
./src/ocean/core/Enforce.d:323 void ocean.core.Enforce.enforceImpl!(">=", ocean.core.Test.TestException, float, int).enforceImpl(float, int, immutable(char)[], int) [0x628138]
/var/lib/jenkins/dlang_projects@2/sociomantic-tsunami/ocean/test/sysstats/main.d:68 int 
```

mihails-strasuns:
Can you run the same pipeline with released 2.077 to see if it is because of compiler or environment?

mihails-strasuns:
Though environment is more likely - https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/test/sysstats/main.d#L68

I guess you have virtualized system where CPU number is not accessible?

mihails-strasuns:
Ping @Burgos - I think you are more familiar with stats module than me.

MartinNowak:
Slaves are using LXD or KVM.

Burgos:
I wonder if the problem is simply because the timing scheduling and CPU counting is different in LXD/KVM than on the bare metal machine. We'll try adding some busy loop there to spend some time in the CPU instead of yielding the control back.

MartinNowak:
`Stats(nan, nan, nan, 24403968, 8146944, 0.024311)`
Reproducibly the uptime diff is 0, while there are as many ticks per second (100), lxcfs rounds them to full seconds.
https://insights.ubuntu.com/2015/03/02/introducing-lxcfs/

Burgos:
Hm, ok. Then we could just sleep for two seconds before getting the results. I don't think this will hurt developers much (and they could always `make -j` to speed the tests). Alternatively, we could just ignore this test on the platforms that behave like this.

MartinNowak:
@Burgos see https://github.com/sociomantic-tsunami/ocean/pull/385.

mihails-strasuns:
Updated to 4.0.0-alpha.2, looking at build log

mihails-strasuns:
Green! D-YAML has failed though :)

Burgos:
Such a time to be alive!

MartinNowak:
> Green! D-YAML has failed though :)

Yes, see https://github.com/dlang/dmd/pull/7315#issuecomment-347937938 :/.

MartinNowak:
Let's hope it doesn't run into any other issues on the KVM host, but since it works on Travis-CI/docker, I wouldn't really expect some.

mihails-strasuns:
@MartinNowak I see. Do you want me to have a look into Jenkins instance taking over what Travis/CircleCI currently do?

MartinNowak:
Let's see whether https://github.com/dlang/ci/pull/90 works.

Burgos:
Am I seeing correctly that merging this PR started failing the master? https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fci/detail/master/47/pipeline

MartinNowak:
Yes @Burgos, looks like it :/.
Maybe there is sth. wrong with the timer wait loop?
https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fci/detail/master/47/pipeline/180#step-313-log-369

Currently rerunning https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fci/detail/master/48/pipeline.

Burgos:
The timer loop looks correct to me. I think the problem is that we're checking for equality of the ProcUptime instances: https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/sys/stats/linux/ProcVFS.d#L209 which includes both `uptime` and `idle` but the CPU user calculation uses only a uptime member. https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/sys/Stats.d#L171

Burgos:
(So the `idle` progresses, but not the `uptime`).

mihails-strasuns:
I think for now best course of action would be to disable this one specific test:

```diff
diff --git a/Build.mak b/Build.mak
index 11eb0f0..f875f76 100644
--- a/Build.mak
+++ b/Build.mak
@@ -20,6 +20,10 @@ clean += .*.lst
 TEST_FILTER_OUT += \
        $C/test/collectd/main.d
 
+# fails with KVM, temporarily disabled
+TEST_FILTER_OUT += \
+       $C/test/sysstats/main.d
+
 $O/test-%: override LDFLAGS += -lebtree
 
 $O/test-filesystemevent: override LDFLAGS += -lrt
```

It will fix pipeline for now and we will talk about best way to fix actual test later this week among us.

Burgos:
https://github.com/sociomantic-tsunami/ocean/pull/386

mihails-strasuns:
https://github.com/dlang/ci/pull/91

PR #90, opened 2017-11-29
use Travis-CI command to test ocean

MartinNowak:
- also add beaver to PATH

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -210,11 +210,10 @@ def testDownstreamProject (name) {
                     break;
 
                 case 'sociomantic-tsunami/ocean':
-                    sh '''
-                    git submodule update --init
-                    make d2conv V=1
-                    make test V=1 DVER=2 F=production ALLOW_DEPRECATIONS=1
-                    '''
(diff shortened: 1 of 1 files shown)


mihails-strasuns:
beaver is a shell utility on top of docker, haven't you mentioned that you don't want to use it for now?

MartinNowak:
I see, read this

> To use Docker in you builds, Beaver expects a Dockerfile

as optional docker support.

mihails-strasuns:
No, sorry for misleading wording. Maybe @leandro-lucarella-sociomantic originally intended it may work both ways thus put it into readme like that, but right now it is tightly tied to docker.

leandro-lucarella-sociomantic:
Yeah, even when beaver is not really tied to docker, I think at the moment all tools use docker. If we are not calling beaver from the build scripts, then there is no point in adding beaver to the path.

Also, I don't know what `test_travis_yaml()` is for, but our travis build uses docker, so if docker is not available here, you probably want to ignore the travis.yml file too.

MartinNowak:
> test_travis_yaml()

Just executes the script of a `.travis.yml`, but nothing else.

PR #98, opened 2017-12-21
parallelize vibe.d tests

MartinNowak:
- using https://github.com/vibe-d/vibe.d/pull/2016

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -158,7 +158,7 @@ def test_travis_yaml () {
 }
 
 def testDownstreamProject (name) {
-    def repo = name // to fix issues with closure
+    def repo = name.tokenize('+')[0]
     node { ws(dir: 'dlang_projects') {
         unstash name: "dlang-build"
         withEnv([
@@ -171,17 +171,19 @@ def testDownstreamProject (name) {
                     // set HOME to separate concurrent ~/.dub user paths
                     "HOME=${env.WORKSPACE}"
                 ]) {
-            try { dir(repo) {
+            try { dir(name) {
 
-                if (repo == 'rejectedsoftware/vibe.d') {
-                    clone("https://github.com/${repo}.git", 'v0.8.3-alpha.1')
+                if (repo == 'vibe-d/vibe.d') {
+                    // https://github.com/vibe-d/vibe.d/pull/2016
+                    // clone("https://github.com/${repo}.git", 'v0.8.3-alpha.1')
+                    clone('https://github.com/MartinNowak/vibe.d.git', 'parallel_tests')
                 } else if (repo == "sociomantic-tsunami/ocean") {
                     clone("https://github.com/${repo}.git", 'v4.0.0-alpha.5')
                 } else {
                     cloneLatestTag("https://github.com/${repo}.git")
                 }
 
-                switch (repo) {
+                switch (name) {
                 case 'gtkd-developers/GtkD':
                     sh 'make DC=$DC'
                     break;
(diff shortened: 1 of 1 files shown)


vars/runPipeline.groovy
```diff
@@ -328,7 +347,13 @@ DFLAGS=-I%@P%/../imports -L-L%@P%/../libs -L--export-dynamic -L--export-dynamic
 
     def dub_projects = [
         // sorted by test time slow to fast
-        "rejectedsoftware/vibe.d",
+        "vibe-d/vibe.d+libevent-base",
+        "vibe-d/vibe.d+libevent-examples",
+        "vibe-d/vibe.d+libevent-tests",
```
wilzbach: Vibe-core is the new default with 0.8.3, so it might be worth a consideration to drop these tests as libevent is on its way out.
After all, they are really expensive and the goal of the project tester is to avoid that we break projects, not add another CI for them.
See: https://github.com/vibe-d/vibe.d/pull/2002

vars/runPipeline.groovy
```diff
@@ -328,7 +347,13 @@ DFLAGS=-I%@P%/../imports -L-L%@P%/../libs -L--export-dynamic -L--export-dynamic
...
+        "vibe-d/vibe.d+vibe-core-base",
+        "vibe-d/vibe.d+vibe-core-examples",
+        "vibe-d/vibe.d+vibe-core-tests",
+        "vibe-d/vibe.d+libasync-base",
```
wilzbach: Imho we should drop this too. At least until [event core](https://github.com/vibe-d/eventcore) supports it...

vars/runPipeline.groovy
```diff
@@ -146,17 +146,19 @@ def testDownstreamProject (name) {
...
                 ]) {
-            try { dir(repo) {
+            try { dir(name) {
 
-                if (repo == 'rejectedsoftware/vibe.d') {
-                    clone("https://github.com/${repo}.git", 'v0.8.3-alpha.1')
+                if (repo == 'vibe-d/vibe.d') {
+                    // https://github.com/vibe-d/vibe.d/pull/2016
+                    // clone("https://github.com/${repo}.git", 'v0.8.3-alpha.1')
+                    clone('https://github.com/MartinNowak/vibe.d.git', 'parallel_tests')
```
wilzbach: You can use upstream now, e.g. by cloning commit `5766e5135b77ab62da21e29ed0660ac9178ea093`

MartinNowak: Clone commit doesn't work.

vars/runPipeline.groovy
MartinNowak: It's quite an important project, so it seems better we cover all the supported code.

MartinNowak:
Looks like this unveiled a dub bug introduced in https://github.com/dlang/dub/pull/1263, will have to track that down first.

wilzbach:
Removed the special casing of Vibe.d as 0.8.3-alpha.3 is now available and rebased to `upstream/master` (without conflicts).

MartinNowak:
Dub bug has been fixed in dlang/dub#1350.

wilzbach:
Looks like there's a new failure coming from of Vibe.d :/
I assume that this is due to Vibe.d switching to using vibe-core by default with the upcoming 0.8.3

```
7E3589D5(----) INF] Receive loop finished.
Receiver watchdog failure.
Program exited with code 1
script returned exit code 2
```

wilzbach:
> I assume that this is due to Vibe.d switching to using vibe-core by default with the upcoming 0.8.3

Or latest tags simply ignores the alpha versions:

```
 > git rev-parse v0.8.2^{commit} # timeout=10
Checking out Revision 38d5da2bd87595cf4c9c6c744301d3ee9300e730 (v0.8.2)
```

(updated the groovy script to an explicit checkout)

wilzbach:
Nice the total time went down from 33m39s to 12m13s!!! Nice job!
For convenience, the run time per component:

component | time
---------------|-----------
vibe.d+libasync-base | 3m45s
vibe.d+libevent-base | 4m20s
vibe.d+libevent-examples | 12m1s
vibe.d+libevent-tests | 8m35s
vibe.d+vibe-core-base | 4m31s
vibe.d+vibe-core-examples | 9m51s
vibe.d+vibe-core-tests | 6m44s

MartinNowak:
Thanks, some of my dub work should significantly reduce the necessary time.

MartinNowak:
> Or latest tags simply ignores the alpha versions

Yes, the regex only matches release versions.

PR #100, opened 2017-12-26
add 2nd GCE on-demand instance

MartinNowak:
- also move instances closer to Jenkins master (FRA)


MartinNowak:
The goal is to replace the permanent instance with a small vHost and only use on-demand instances for testing, as the server is mostly idling.

PR #102, opened 2017-12-27
Add more projects from dlang-community

wilzbach:
Sadly not all projects use the same setup and respect $DC.

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -230,17 +230,6 @@ def testDownstreamProject (name) {
                     '''
                     break;
 
-                case 'dlang-community/libdparse':
-                    sh 'cd test && DMD=$DC ./run_tests.sh'
-                    break;
-
-                case 'dlang-community/dcd':
-                    sh '''
-                    git submodule update --init --recursive
-                    make test -j2 DMD=$DC
-                    '''
-                    break;
-
                 default:
                     test_travis_yaml()
                     break;
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -169,10 +169,6 @@ def testDownstreamProject (name) {
                 }
 
                 switch (repo) {
-                case ['Hackerpilot/DCD', 'Hackerpilot/dfix']:
-                    sh 'make DMD=$DMD'
-                    break;
-
                 case 'gtkd-developers/GtkD':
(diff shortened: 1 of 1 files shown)


vars/runPipeline.groovy
```diff
@@ -344,13 +351,16 @@ DFLAGS=-I%@P%/../imports -L-L%@P%/../libs -L--export-dynamic -L--export-dynamic
...
+        "BlackEdder/ggplotd", // 1m56s
+        "dlang-community/D-Scanner", // 1m40s
+        "rejectedsoftware/diet-ng", // 56s
+        "atilaneves/unit-threaded", //36s
+        "d-gamedev-team/gfm", // 28s
+        "dlang-community/DCD", // 23s
+        "CyberShadow/ae", // 22s
+        "dlang-community/D-YAML", // 15s
+        "dlang-community/libdparse", // 13s
+        "dlang-community/dfmt", // 11s
```
MartinNowak: We should prolly start to write the times to a persisted json file and read this dynamically.
https://jenkins.io/doc/pipeline/steps/workflow-basic-steps/#code-unarchive-code-copy-archived-artifacts-into-the-workspace

vars/runPipeline.groovy
```diff
@@ -234,6 +230,17 @@ def testDownstreamProject (name) {
                     '''
                     break;
 
+                case 'dlang-community/libdparse':
+                    sh 'cd test && DMD=$DC ./run_tests.sh'
```
MartinNowak: We do expose both DMD and DC in the pipeline, just as the install.sh script does.

vars/runPipeline.groovy
```diff
@@ -234,6 +230,17 @@ def testDownstreamProject (name) {
...
+                    break;
+
+                case 'dlang-community/dcd':
+                    sh '''
+                    git submodule update --init --recursive
```
MartinNowak: The travis.sh script does sth. similar, sure just running the travis.yml script fails?

wilzbach:
>  I'll try to run them with their verbatim travis.yml script though.

I wasn't sure whether the `dmd` binary will point to the freshly built one as these two ignore the environment variables.

MartinNowak:
Yes, there is no system-wide dmd on the testers.

PR #105, opened 2018-01-02
Try to fix 'dub search failed'

wilzbach:
https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-7576/5/pipeline

```
[ERROR] /var/lib/jenkins/dlang_projects@2/dlang/dub/test/feat663-search.sh:11 `dub search dub` failed
        Using dub registry url 'https://code-mirror.dlang.io/'
        Refreshing local packages (refresh existing: true)...
        Looking for local package map at /var/lib/dub/packages/local-packages.json
        Looking for local package map at /var/lib/jenkins/dlang_projects@2/.dub/packages/local-packages.json
        Try to load local package map at /var/lib/jenkins/dlang_projects@2/.dub/packages/local-packages.json
        Determined package version using GIT: dub 1.6.0
        ==== registry at https://code-mirror.dlang.io/ (fallback registry at https://code-mirror.dlang.io/) ====
        dub (1.7.0-rc.1)                Package manager for D packages
```

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -196,6 +196,7 @@ def testDownstreamProject (name) {
                 case 'dlang/dub':
                     sh '''
                       rm test/issue884-init-defer-file-creation.sh # FIXME
+					  rm test/feat663-search.sh # FIXME
                       sed -i \'s| defaultRegistryURL = .*;| defaultRegistryURL = "https://code-mirror.dlang.io/";|\' source/dub/dub.d
                       sed -i \'/^source.*activate/d\' travis-ci.sh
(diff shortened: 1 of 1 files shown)


MartinNowak:
Is this test still flaky? I've split the search the grep step to better localize the error.
https://github.com/dlang/dub/pull/1324
Also see #106 which uses dub's default fallback mechanism, atm. the Jenkins runners only use code-mirror.dlang.io which suffers from the same regular OOM restarts as code.dlang.org.
Let's try to go with those for now instead of bluntly disabling the test.

PR #110, opened 2018-01-05
HTTP/2 frontend

MartinNowak:
- support HTTP/2 on ci.dlang.io and code-mirror.dlang.io
- also fix weak DH param


MartinNowak:
Slow (30s) POSTs in haproxy 1.8.3, apparently fixed in upcoming 1.8.4.

PR #111, opened 2018-01-06
temporarily remove unit-threaded until failure can be investigated

JinShil:
This is preventing some PRs from being merged.

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -207,6 +207,12 @@ def testDownstreamProject (name) {
                     sh 'DC=dmd VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0 ./travis-ci.sh || echo failed' // FIXME
                     break;
 
+                case 'atilaneves/unit-threaded':
+                    // FIXME: temporarily remove failing unit-threaded tests until they can be investigated
+                    // see: https://github.com/dlang/ci/pull/111
+                    sh './test.sh || echo failed' // FIXME
+                    break;
+
                 case 'rejectedsoftware/diet-ng':
                     sh 'sed -i \'/mkdir build && cd build/,//d\' .travis.yml' // strip meson tests
                     test_travis_yaml()


wilzbach:
CC @atilaneves

wilzbach:
Example: https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-7613/1/pipeline

wilzbach:
This is quite crude. Would you mind doing it in a similar way like Vibe.d does? I mean the ` || echo failed` - the advantage is that one still sees the error message on the CI.

And now that you have a PR number, how about referencing it in the sources?

(I currently on the go - otherwise I would have made those tiny changes myself)

JinShil:
Sorry @wilzbach, but I really don't understand.  If I follow the vibe.d example, I think I can remove just the failing tests.  I used the jenkins output to identify the failing tests and added a shell script just like vibe.d to remove the failures.  I hope that's what you mean.  If not, please elaborate.  I'm sorry, but I really don't understand how the logic in this .groovy file works.

vars/runPipeline.groovy
```diff
@@ -207,6 +207,16 @@ def testDownstreamProject (name) {
...
                     break;
 
+                // FIXME
+                // temporarily remove failing unit-threaded tests until they can be investigated
+                case 'atilaneves/unit-threaded':
+                    sh 'rm -rf tests/module_with_attrs.d'
+                    sh 'rm -rf tests/module_with_setup.d'
+                    sh 'rm -rf tests/module_with_tests.d'
+                    sh 'rm -rf tests/parametrized.d'
+                    sh 'rm -rf tests/tags.d'
```
wilzbach: You remove file that don't exist (https://github.com/atilaneves/unit-threaded/tree/master/tests) and don't even execute the testsuite here.

atilaneves:
This has already been fixed in unit-threaded.

wilzbach:
@atilaneves: the Jenkins project tester checks out the latest tag by default.
If that has been fixed, it's great. Would you be so kind to open a PR and remove the workaround. I have a feeling that you are the best person for this as it will be easy for you to take actions in case something still doesn't work.

atilaneves:
@wilzbach On it.

Issue #112, opened 2018-01-06
Updates sometimes don't propogate


wilzbach:
PR: https://github.com/dlang/dmd/pull/7553

was fixed in https://github.com/dlang/ci/pull/108 and even though it was repeatedly restarted afterwards, e.g.

https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-7553/7/pipeline/183

It still showed;

> Checking out Revision 2427af70404bbc8878da7295d0e662a458a6a93e (v4.0.0-alpha.2)

Found by @JinShil

wilzbach:
Rebasing the PR to `upstream/master` fixed the problem: https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-7553/8/pipeline/186

ibuclaw:
It looks like its more of a timing issue, and just pot luck depending on which runner you hit.

Looking at the build history of the last 2 hours (Sunday, Jan 7, 12pm UTC)
https://ci.dlang.io/view/all/builds

Timeline is this:

Time (approx) | Change | Runner | Result
------------ | ------------ | ------------- | -------------
10:25 | D Programming Language » ci » PR-115 # 1 | runner-01.ci.lxd | Pass
10:45 | D Programming Language » ci » master # 66 | runner-02.ci.lxd | Pass
10:51 | D Programming Language » dmd » PR-7625 # 2 | runner-01.ci.lxd | Pass
10:52 | D Programming Language » dmd » PR-7624 # 2 | runner-02.ci.lxd | Fail
10:54 | D Programming Language » dmd » PR-7622 # 2 | runner-02.ci.lxd | Fail
10:54 | D Programming Language » dmd » PR-7620 # 2 | runner-02.ci.lxd | Fail
10:55 | D Programming Language » dmd » PR-7619 # 2 | runner-02.ci.lxd | Fail
11:03 | D Programming Language » dmd » PR-7617 # 13 | runner-01-gce | Pass
11:24 | D Programming Language » dmd » PR-7619 # 3 | runner-01.ci.lxd | Pass
11:25 | D Programming Language » dmd » PR-7620 # 3 | runner-02.ci.lxd | Pass

The first two are for a commit that should have fixed all builds.

The first successful build on `runner-02.ci.lxd` was around an hour later.

Looking at the logs of the failed builds, i.e: https://ci.dlang.io/job/dlang-org/job/dmd/job/PR-7619/2/consoleFull

[...]

PR #115, opened 2018-01-07
Temporarily patch Ocean to workaround missing out contract patch

wilzbach:
The patch applied in https://github.com/sociomantic-tsunami/ocean/pull/427
doesn't seem to appear in Ocean `v4.0.0-alpha.4` anymore.

Currently it fails:

https://github.com/dlang/dmd/pull/7617

However, Jenkins did pass on
https://github.com/dlang/dmd/pull/7553 and e.g. on
https://github.com/dlang/phobos/pull/6002

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -246,8 +246,11 @@ def testDownstreamProject (name) {
                     break;
 
                 case 'sociomantic-tsunami/ocean':
+                	// FIXME: https://github.com/sociomantic-tsunami/ocean/pull/427 is only applied for 3.x.x
+                	// See also: https://github.com/dlang/dmd/pull/7617
                     sh '''
                     git submodule update --init
+                    sed "/this.outer.occupied = false;/d" -i src/ocean/net/http/HttpResponse.d # FIXME
                     make d2conv V=1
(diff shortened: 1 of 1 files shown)


wilzbach:
See also: https://github.com/dlang/ci/issues/112

CC mihails-strasuns-sociomantic as looking at:

https://github.com/sociomantic-tsunami/ocean/blob/v4.0.0-alpha.4/src/ocean/net/http/HttpResponse.d#L466
https://github.com/sociomantic-tsunami/ocean/blob/v4.x.x/src/ocean/net/http/HttpResponse.d

still shows the bogus contract.

wilzbach:
The git commit sha checked out is the same: `0cf1550792992f4d868e0ae7fe67b4025790c40d` at dlang/dmd#7553  that introduced the check and passed Jenkins compared with the PRs submitted later which fail, e.g. dlang/dmd#7617

mihails-strasuns-sociomantic:
Beg my pardon - I have forgot to make a point release with the fix before propagating changes to v4.x.x, thus it wasn't present. Will fix in a moment.

mihails-strasuns-sociomantic:
https://github.com/dlang/ci/pull/116

PR #119, opened 2018-01-12
Add dlang-tour to the tested projects

wilzbach:
It regularily broke in the past..

Broken build: https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fci/detail/PR-119/1/pipeline/200

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -251,6 +251,11 @@ def testDownstreamProject (name) {
                     sh 'make test DCOMPILER=$DC'
                     break;
 
+                case 'dlang-tour/core':
+                    sh 'git submodule update --init public/content/en'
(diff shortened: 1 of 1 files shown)


vars/runPipeline.groovy
```diff
@@ -376,6 +381,8 @@ def call() { timeout(time: 1, unit: 'HOURS') {
         "dlang-community/D-YAML", // 15s
         "dlang-community/libdparse", // 13s
         "dlang-community/dfmt", // 11s
+        // not known yet
```
wilzbach: Reminder to myself: this is 1m17s -> needs to be updated

wilzbach:
DLang-Tour is a more or less official project by dlang, so I doubt that @MartinNowak minds about adding it to the projects tested.

Issue #120, opened 2018-01-14
Always checks out upstream/master (instead of the PR branch)


wilzbach:
tl;dr: https://github.com/dlang/dmd/pull/7688 should have shown an CI error.
It's only displayed on the builds now

```
src/zmqd.d(462,40): Error: `data.ptr` cannot be used in `@safe` code, use `&data[0]` instead
src/zmqd.d(477,40): Error: `data.ptr` cannot be used in `@safe` code, use `&data[0]` instead
src/zmqd.d(567,46): Error: `data.ptr` cannot be used in `@safe` code, use `&data[0]` instead
src/zmqd.d(582,46): Error: `data.ptr` cannot be used in `@safe` code, use `&data[0]` instead
src/zmqd.d(624,52): Error: `data.ptr` cannot be used in `@safe` code, use `&data[0]` instead
src/zmqd.d(636,52): Error: `data.ptr` cannot be used in `@safe` code, use `&data[0]` instead
src/zmqd.d(1662,54): Error: `value.ptr` cannot be used in `@safe` code, use `&value[0]` instead
src/zmqd.d(840,66): Error: template instance zmqd.Socket.setArrayOption!() error instantiating
src/zmqd.d(1892,9): Error: `items.ptr` cannot be used in `@safe` code, use `&items[0]` instead
```

https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdruntime/detail/PR-2018/7/pipeline

On https://github.com/dlang/dmd/pull/7688  the log is:

```
> git rev-parse --is-inside-work-tree # timeout=10
Fetching changes from the remote Git repository
 > git config remote.origin.url https://github.com/dlang/dmd.git # timeout=10
Cleaning workspace
 > git rev-parse --verify HEAD # timeout=10
Resetting working tree
 > git reset --hard # timeout=10
 > git clean -fdx # timeout=10
Fetching upstream changes from https://github.com/dlang/dmd.git
 > git --version # timeout=10
 > git fetch --tags --progress https://github.com/dlang/dmd.git +refs/heads/*:refs/remotes/origin/*
 > git rev-parse origin/master^{commit} # timeout=10
Checking out Revision b1f6aec796286a08395d689b54dacdd01d141f9e (origin/master)
 > git config core.sparsecheckout # timeout=10
 > git checkout -f b1f6aec796286a08395d689b54dacdd01d141f9e
Commit message: "Merge pull request #7695 from rainers/default_platform_toolset"
[...]

PR #121, opened 2018-01-14
Clone zmq due to #120

wilzbach:
I will look into submitting a PR to zmq soon.


wilzbach:
Oh that has already been fixed in master: https://github.com/kyllingstad/zmqd/pull/21

PR #122, opened 2018-01-14
use testee's origin URL as fallback for missing env.CHANGE_URL 

wilzbach:


wilzbach:
So `env.CHANGE_URL` isn't defined:

image

MartinNowak:
See https://github.com/dlang/ci/pull/33 for the true source link.

MartinNowak:
There is also https://issues.jenkins-ci.org/browse/JENKINS-40876, but that should be fixed.

MartinNowak:
getKey looks useful https://github.com/jenkinsci/git-plugin/blob/7a2e2f126c6767ac488907288451545dc5ca8035/src/main/java/hudson/plugins/git/GitSCM.java#L1695

MartinNowak:
Exactly what we need, let me put together the fix.

vars/runPipeline.groovy
```diff
@@ -291,10 +278,21 @@ def buildDlang() {
...
+        // workaround unset CHANGE_URL by using GitSCM's RemoteConfig, which contains testee's origin URI.
+        // see dlang/ci#122
+        //
+        // match either
+        // https://github.com/dlang/dmd/pull/123 (env.CHANGE_URL)
+        // or
+        // https://github.com/dlang/dmd.git (scm origin URI)
+        def remote_url = env.CHANGE_URL ?: scm.repositories[0].URIs[0].toString()
+        def testee = (remote_url =~ /github.com\/[^\/]+\/([^\/\.]+)/)[0][1] // e.g. dmd
+        echo "Used ${env.CHANGE_URL ? 'CHANGE_URL' : 'scm repo URI'} '${remote_url}' to determine '${testee}' as testee."
```
wilzbach: > Used scm repo URI 'https://github.com/dlang/ci.git' to determine 'ci' as testee.

PR #130, opened 2018-01-24
Add dlang/tools to the tested projects

wilzbach:
I just realized we don't run the testsuite of `dlang/tools`.

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -232,6 +232,12 @@ def testDownstreamProject (name) {
                     sh 'DC=$DC ./travis-ci.sh'
                     break;
 
+                case 'dlang/tools':
+                    // explicit test to avoid Digger setup, see dlang/tools#298 and dlang/tools#301
+                    sh "make -f posix.mak all DMD='${env.WORKSPACE}/distribution/bin/dmd'"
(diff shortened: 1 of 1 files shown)


vars/runPipeline.groovy
```diff
@@ -387,6 +387,7 @@ def call() { timeout(time: 1, unit: 'HOURS') {
         "dlang-community/D-YAML", // 15s
         "dlang-community/libdparse", // 13s
         "dlang-community/dfmt", // 11s
+        "dlang/tools", // not known yet
```
wilzbach: It takes quite a while at the moment - 2m44s:

image

However, this is due to using digger which builds `dmd`, `druntime` and `phobos` from scratch. Once this PR is merged, we don't need to do this as the tools's testsuite will be run on every PR to dmd, druntime and tools. So we can remove the entire digger setup at `tools` and then update the time. I suspect it's somewhere between 30-60s.

wilzbach: tools PR: https://github.com/dlang/tools/pull/298

MartinNowak:
Seems blocked until we have a tagged version of dlang/tools that no longer uses Digger in it's test.
Not really interested in 8 * 500 MiB storage just for digger.

PR #139, opened 2018-01-25
Try to nuke ~/.dub once

wilzbach:
Required for https://github.com/dlang/phobos/pull/5921

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -358,6 +358,9 @@ def buildDlang() {
             sh '''#!/usr/bin/env bash
             set -ueo pipefail
 
+			# Nuking ~/.dub is required to always fetch the latest package
+            rm -rf $HOME/.dub
+
             rm -rf distribution
             mkdir -p distribution/{bin,imports,libs}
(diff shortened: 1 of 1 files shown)


wilzbach:
On a second thought. We already use a local mirror, so let's see whether this adds a noticeable time to Jenkins, it's quite likely that we run into issues with this again & either doing a PR to nuke `~/.dub`  (or waiting for the worker to die and be "resurrected") doesn't sound like the best idea.

MartinNowak:
The was not related to fetched ddox packages last time I debugged sth. similar. It was rather a hardcoded fetch in dub's test-suite or so.

And no deleting `~/.dub` is not OK, packages have to be re-downloaded from **github**.
That's also the wrong place to delete `.dub` (it's on the node that builds dlang), we use one `.dub` folder per testee.

wilzbach:
Sorry - it was just an experiment.

MartinNowak:
No worries, just want to let you know why we cache things.

MartinNowak:
I've manually deleted ddox-* on all slaves.

PR #141, opened 2018-01-25
Disable ggplotd

wilzbach:
It's always running into timeouts lately. So I will disable it for now and then open a revert PR in which this can be looked at in more detail.


wilzbach:
Ah it seems that this only appears for https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-6676/13/pipeline aka dlang/dmd#6676 and I can reproduce the timeout locally :(

PR #142, opened 2018-01-26
optimize test order for minimal queue length

MartinNowak:
- rather than for overall delay

- the many slow test-suites immediately clog all executors on a single PR

- mainly to prevent booting an on-demand instance just to quickly run
  the remaining queued tests

- will slightly increase overall latency a bit, as slow tests are only
  started after all fast tests are done


wilzbach:
(rebased as DStep has been added)

Issue #135, opened 2018-01-25
Jenkins's fails to write files (locking problem?)


wilzbach:
```
[rejectedsoftware/vibe.d] vibe-d:textfilter 0.8.3-alpha.1: building configuration "library"...
[rejectedsoftware/vibe.d] ../../core/vibe/core/log.d(565,21): Deprecation: Built-in hex string literals are deprecated, use `std.conv.hexString` instead.
[rejectedsoftware/vibe.d] vibe-d:inet 0.8.3-alpha.1: building configuration "library"...
[rejectedsoftware/vibe.d] ../../core/vibe/core/log.d(565,21): Deprecation: Built-in hex string literals are deprecated, use `std.conv.hexString` instead.
[rejectedsoftware/vibe.d] vibe-d:tls 0.8.3-alpha.1: building configuration "openssl"...
[rejectedsoftware/vibe.d] ../../core/vibe/core/log.d(565,21): Deprecation: Built-in hex string literals are deprecated, use `std.conv.hexString` instead.
[rejectedsoftware/vibe.d] vibe-d:http 0.8.3-alpha.1: building configuration "library"...
[rejectedsoftware/vibe.d] ../../core/vibe/core/log.d(565,21): Deprecation: Built-in hex string literals are deprecated, use `std.conv.hexString` instead.
[rejectedsoftware/vibe.d] immutable(ubyte)[]
[rejectedsoftware/vibe.d] ../../http/vibe/http/server.d(2026,56): Deprecation: variable vibe.http.server.HTTPServerOption.distribute is deprecated - Use runWorkerTaskDist or start threads separately. It will be removed in 0.9.
[rejectedsoftware/vibe.d] Error: Error writing file '../../http/.dub/build/library-debug-linux.posix-x86_64-dmd_2078-BD83D681FC3A568855FE06D6DCD0FFE1/libvibe-d_http.a'
[rejectedsoftware/vibe.d] dmd failed with exit code 1.
```

https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdruntime/detail/PR-2061/1/pipeline

wilzbach:
Happened here again: https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-7769/4/pipeline

wilzbach:
Or here: https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-7768/5/pipeline

[...]

PR #145, opened 2018-01-28
Add DStep to the tested projects

jacob-carlborg:


jacob-carlborg:
Why is it checking out the latest tag? I want it to test master.

jacob-carlborg:
I looked down the clone to the current master. I'll remove this later when a new tag is added for DStep.

wilzbach:
It looks like the Jenkins Git checkout plugin only supports tags and branches :O

jacob-carlborg:
Added a tag.

jacob-carlborg:
It's Dub that is failing.

wilzbach:
Yeah that's unrelated and due to https://github.com/dlang/ci/issues/135

However, I can't overwrite the CI check here. I will try to restart it.

MartinNowak:
Downloading a 200+MiB clang+llvm for every test is **not OK**, please don't unnecessarily waste resources, simply because they are not your resources.
https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fci/detail/PR-146/2/pipeline/207#step-472-log-8
I've installed [libclang-4.0-dev](https://packages.ubuntu.com/xenial-updates/libclang-4.0-dev) on all testers. I'm reverting *this* until dstep manages to detect that package.
https://github.com/dlang/ci/pull/148
Let us know if you need sth. else installed.

You should also fix your Travis-CI tests to either use packages or at least cache the downloads, it's currently downloading 13x200+MiB from the llvm site.
https://travis-ci.org/jacob-carlborg/dstep/builds/334364953

MartinNowak:
FYI @jacob-carlborg

jacob-carlborg:
> You should also fix your Travis-CI tests to either use packages or at least cache the downloads, it's currently downloading 13x200+MiB from the llvm site.
https://travis-ci.org/jacob-carlborg/dstep/builds/334364953

Should I cache that to reduce the traffic to the LLVM site? Because the Travis documentation says:

"Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source:" [1]

[1] https://docs.travis-ci.com/user/caching/#Things-not-to-cache

jacob-carlborg:
> You should also fix your Travis-CI tests to either use packages.

I cannot use packages because they are too old. I want to control myself which versions of LLVM I support.

PR #151, opened 2018-01-29
Temporarily disable ggplotd

wilzbach:
See also: https://github.com/BlackEdder/ggplotd/pull/45


andralex:
thx

timotheecour:
should revert once https://github.com/dlang/dmd/pull/7930 is pushed

PR #155, opened 2018-02-01
bump slave disk size to 40GiB

MartinNowak:


wilzbach:
Shouldn't we nuke the entire project after running it?
I only see this in the last step:

```sh
dub clean --all-packages >/dev/null
# workaround https://github.com/dlang/dub/issues/1256
if [ -d '${env.WORKSPACE}/.dub/packages' ]; then
    find '${env.WORKSPACE}/.dub/packages' -type f -name '*.a' -delete
fi
git -C '${name}' clean -dxf >/dev/null
rm -r '${env.WORKSPACE}/distribution'
```

PR #163, opened 2018-02-04
Remove more spurious, flaky vibe.d tests for now

wilzbach:


MartinNowak:
Thanks, it's rather devastating for a test-suite though. At least of 'em are filed upstream.

PR #164, opened 2018-02-04
Test tools with MODEL=32 too

wilzbach:


vars/runPipeline.groovy
```diff
@@ -238,6 +238,10 @@ def testDownstreamProject (name) {
                     // explicit test to avoid Digger setup, see dlang/tools#298 and dlang/tools#301
                     sh "make -f posix.mak all DMD='${env.WORKSPACE}/distribution/bin/dmd'"
                     sh "make -f posix.mak test DMD='${env.WORKSPACE}/distribution/bin/dmd' DFLAGS="
+                    // test 32-bit build of the D tools
+                    sh "make -f posix.mak clean"
+                    sh "make -f posix.mak all MODEL=32 DMD='${env.WORKSPACE}/distribution/bin/dmd'"
+                    sh "make -f posix.mak test MODEL=32 DMD='${env.WORKSPACE}/distribution/bin/dmd' DFLAGS="
```
WebDrake: Is there a reason for not explicitly marking the existing tests with `MODEL=64` ... ?  Or is the assumption that the setup as written will be more robust to being run in both 32- and 64-bit host environments?

PR #165, opened 2018-02-07
Add automem to the list of tested projects

wilzbach:
CC @atilaneves

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -423,6 +423,7 @@ def call() { timeout(time: 1, unit: 'HOURS') {
         "Netflix/vectorflow",
         "PhilippeSigaud/Pegged",
         "ariovistus/pyd",
+        "atilaneves/automem",
         "dlang/undeaD",
         "economicmodeling/containers",
         "kyllingstad/zmqd",


atilaneves:
The failure seems to be trying to build an old version of automem for some reason. I just pushed a new one just in case.

wilzbach:
@atilaneves the project tester will always use the latest, stable git tag. I have force-pushed here to see restart the CI. Maybe that helps ;-)

atilaneves:
Ah, I got confused by the error message - it's failing in `preBuildCommands` using an old version of unit-threaded. I hate that dub does that with dub run.

I just pushed another version to master, if you could somehow force the jenkins CI it should be green now.

wilzbach:
> I hate that dub does that with dub run.

You know where to send a PR to ;-)

>  if you could somehow force the jenkins CI it should be green now.

force-pushed.

atilaneves:
@wilzbach Doesn't seem to have updated.

wilzbach:
I see the following:

```
unit-threaded 0.7.24: building configuration "library"...
```

Fair enough, but that's valid according to your dependency file:

```
            "dependencies": {
                "unit-threaded": "~>0.7.0",
           }
```

I suggest bumping the `dub.json`, tagging a new version and pinging me here. I'm happy to restart.

atilaneves:
There's a dub.selections.json checked in though - it should never pick up an older version. The Travis build is green, for instance.

wilzbach:
> There's a dub.selections.json checked in though - it should never pick up an older version. 

https://github.com/atilaneves/automem/blob/v0.0.10/dub.selections.json

> The Travis build is green, for instance.

You test with master, not the last git tag ;-)

atilaneves:
@wilzbach I must have been asleep when I read this thread because I thought it was master. I pushed a new tag.

wilzbach:
No worries. Force-pushed.

atilaneves:
The failures are real now. I had to install the nightly dmd to reproduce. dmd 2.078.2 works fine.

It might be a good idea to merge anyway even if red - compiler changes broke previously working code.

wilzbach:
> It might be a good idea to merge anyway even if red - compiler changes broke previously working code.

That's why we add it here ;-)
However, I strongly object to breaking the CI on purpose. 

FYI: the failures are expected and aren't a regression. They are do to changes in experimental, which is allowed to break.

https://github.com/dlang/phobos/pull/5921
https://forum.dlang.org/post/730a99d2-1043-19b7-551a-c6f57790f422@erdani.org

tl;dr: use the `stdx-allocator` package

atilaneves:
Just pushed v0.1.0. Should be green.

wilzbach:
Thanks! Force-pushed again. Let's see ;-)

wilzbach:
image

Thanks!
I will wait a bit for @MartinNowak to see this PR and approve it.

wilzbach:
Okay I'm going to merge this now as imho `automem` is an important project in terms of D's slogan of "fast code fast". Also it's just a tiny library, so it hopefully won't cause us any problems and if it does, then well, we most likely found a regression ;-)

PR #175, opened 2018-02-13
Add mir projects to the Project Tester

wilzbach:
This should prevent regressions like https://issues.dlang.org/show_bug.cgi?id=18430

CC @9il

FYI: the latest stable git tag is checked out

(this probably first requires the regression fix)

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -303,6 +303,14 @@ def testDownstreamProject (name) {
                     sh 'dub build -c std'
                     break;
 
+                case 'libmir/mir-algorithm':
+                    sh 'dub test --compiler=$DC'
+                    break;
+
+                case 'libmir/mir':
+                    sh 'dub test --compiler=$DC'
+                    break;
+
                 default:
                     test_travis_yaml()
                     break;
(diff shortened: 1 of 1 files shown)


wilzbach:
Yeah it breaks as expected :/

image

vars/runPipeline.groovy
```diff
@@ -445,6 +453,9 @@ def call() { timeout(time: 1, unit: 'HOURS') {
         "nomad-software/dunit",
         "repeatedly/mustache-d",
         "s-ludwig/taggedalgebraic",
+        "libmir/mir-algorithm",
+        "libmir/mir-random",
```
9il: What is the difference between mir-random and others that are in the switch?

wilzbach: The Project Tester will try the `script` line of the `.travis.yml` by default.
That's generally preferred, because that's puts less maintenance effort here.
However, for `mir-algorithm` there would be no `travis_wait` and for `mir` we don't want to fetch `imageformats` just to test the examples (at least not while DUB is still moderately unstable).

Does this answer your question?

9il:
Yes, thank you

wilzbach:
Force-push to restart Jenkins - it should pass now (:

wilzbach:
And another rebase now that the rough compilation times are known.

I'm going to put my hand in the fire for these additions because:

- The mir projects are very important to D and its mantra of "Fast code fast"
- It repeatedly ran into breakage and it's a lot better to catch the regression before they get into master
- I have merge access to all three projects in case something goes wrong
- I have a personal interest in these projects
- They don't do any crazy things that could lead to random failures (network requests, multi-threading, event loops, ...)

wilzbach:
Not sure. Can we build it without external libraries?
If so, go ahead and open a PR :)

PR #183, opened 2018-03-01
Add DWT to the tested projects

jacob-carlborg:


jacob-carlborg:
https://github.com/dlang/dmd/pull/7969 needs to be merged into master before this is passing.

jacob-carlborg:
@wilzbach everything is green now.

PR #185, opened 2018-03-12
Build DMD, Druntime + Phobos in debug mode

wilzbach:
Not sure whether it's a good idea as this might result in a slowdown.
Also in theory there might be differences between the debug and release version and thus could result in an undetected breakage.

OTOH seeing a nice stacktrace on errors at the CI helps, though of course it can be argued that this is needed so seldom that it's not worth the risk or slowdown.


MartinNowak:
Somewhat undecided.

This goes back to building dmd with assertions but without the invariant slowdown.
https://github.com/dlang/dmd/pull/7980
Though the test suite doesn't spend so much time building code I guess.
Another issue is the up&download size of the binary artifacts, hopefully compressions works well enough here.

PR #189, opened 2018-03-30
Build dub in gdb to get the stacktrace

wilzbach:
Trying to to get the root cause of https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-6391/2/pipeline/


wilzbach:
Urgh that is passing. So I'm merging this in the hope that if the failure appears again, we get a proper stack trace ...

Issue #190, opened 2018-03-31
Non-optional dependency diet-ng of vibe-d:http not found in dependency tree!?.


wilzbach:
This happens very often now on the Jenkins CI.

Related issues
- https://github.com/dlang/dub/issues/1375
- https://github.com/dlang/dub/issues/1180
- https://github.com/dlang/dub/pull/1148
- https://github.com/dlang/dub/issues/1112
- https://github.com/dlang/dub/issues/986
- https://github.com/dlang/dub/issues/865

Maybe it's a regression of https://github.com/dlang/dub/issues/828 / https://github.com/dlang/dub/issues/803? 

Example: https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-5822/6/pipeline

wilzbach:
So this definitely looks a dub regression, both openssl nor botan had changes in the last weeks:

https://github.com/etcimon/botan/blob/master/dub.json#L118
http://code.dlang.org/packages/openssl

wilzbach:
More examples:
- https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-6400/7/pipeline
- https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-6396/3/pipeline

> Fetching openssl 1.1.6+1.0.1g (getting selected version)...
> ...
> Non-optional dependency openssl of botan not found in dependency tree!?.

Also in diet-ng (with openssl too)

- https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-6391/7/pipeline
- https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-5822/6/pipeline/
- https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-5822/6/pipeline

Also in vibe-d+libevent:
- https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-6388/13/pipeline

```
+ cd examples/app_skeleton
+ dub build --compiler=dmd --override-config=vibe-d:core/libevent --build-mode=separate
Fetching diet-ng 1.4.5 (getting selected version)...
Fetching openssl 1.1.6+1.0.1g (getting selected version)...
```

wilzbach:
Or in dlang-requests:

```
+ dub build -c std
[...]

PR #191, opened 2018-04-01
Temporarily disable projects depending on Vibe.d due to #190

wilzbach:


wilzbach:
So we can't leave our CI on red forever. It has been red for almost a week!

Hence, I went ahead and temporarily disabled the failing projects.
Once the cause of #190 has been identified, this should be reverted.

PR #213, opened 2018-07-05
Add Mecca to the Project Tester

wilzbach:
CC @Shachar, @tomerfiliba, @JohanEngelen I noticed that 2.081 broke building Mecca and that's very unfortunate.
With enabling it at the project tester we can guarantee that all future PRs need to pass Mecca's testsuite.
The latest git tag is checked out and by default the testing instructions from the `.travis.yml` are performed.


wilzbach:
Of course as Mecca currently doesn't build with 2.081, this will need https://github.com/weka-io/mecca/pull/6 and a new git tag to pass the CI.

PR #218, opened 2018-07-06
Add dagon to the Project Tester

wilzbach:
CC @gecko0307 are you okay with having a bit more testing of your projects?
For this to work, I think you need to update dlib, s.t. it doesn't use the removed `std.c` modules.
Then in the future, such hard breakage can't happen again.

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -136,6 +136,7 @@ projects=(
     "BBasile/iz" # 12s
     "aliak00/optional" # 12s
     "dlang-community/dfmt" # 11s
+    "gecko0307/dagon" # TODO
     # run in under 10s sorted alphabetically
     "Abscissa/libInputVisitor"
     "ariovistus/pyd"
--- a/buildkite/load_distribution.sh
+++ b/buildkite/load_distribution.sh
@@ -22,3 +22,6 @@ export TEMP=$PWD/tmp
 
 export DC=dmd
 export DMD=dmd
+
+# At the moment all workers are x86_64
+export ARCH=x86_64


gecko0307:
@wilzbach if you mean `std.c` error with latest DMD, then it's fixed (https://github.com/gecko0307/dlib/commit/ba8e0b9dbdece5e975cd166da1c1decc5f1f4196). It's not released yet, so Dagon doesn't compile with 2.081.0 right now.

wilzbach:
Yep, I was referring to this. Could you give me a quick ping once its released?

gecko0307:
Sure.

gecko0307:
Now Dagon should compile with DMD 2.081.0 and LDC 1.10.0.

gecko0307:
Are only releases tested there? I'm asking because master branch now works, but there's still a lot of work to be done until the next version of Dagon.

wilzbach:
> Are only releases tested there?

Yes, the latest git tag is tested here. However, it currently fails as the `dub.selections.json` file has priority over the `dub.json` file and in `dub.selections.json` you still reference dlib 0.13
Thus I get the same error when I run `dub test` locally.

>  but there's still a lot of work to be done until the next version of Dagon.

Well, the idea of the project tester is that once added a new DMD release doesn't break your testsuite anymore (or each individual PR to be exact).
All that matters to is that you don't `git tag` a new release that fails your testsuite, that's the only big condition for projects being here ;-)
That being said, if you want to wait to the next version of Dagon, that would be totally fine for me.

ZombineDev:
Thanks for the ping, I started a new build.

wilzbach:
```
Error processing arguments: Missing value for argument --arch=.
```

It looks like either we should set `ARCH` or need a custom build command here. Probably just setting `ARCH` might be easier though.

wilzbach:
> Probably just setting ARCH might be easier though.

Did this in https://github.com/dlang/ci/pull/218/commits/a1968a4f181bf60bb59789be7c573eb660117232

PR #224, opened 2018-07-07
Add intel-intrinsics package

wilzbach:
CC @p0nce - ping me once it's passing on 2.081 and then we can get this merged here.


p0nce:
Fixed https://github.com/AuburnSounds/intel-intrinsics
The test exercise DMD only a little (SIMD emulation instead of D_SIMD, and no -O) but will be made stronger in the future.

p0nce:
ping @wilzbach

PR #225, opened 2018-07-08
Add buildkite config script

wilzbach:
Preview: https://buildkite.com/d-lang/dmd-test/builds
TODO:
- [ ] create pipeline for each repository and add webhook for dmd,druntime,phobos,tools,dub (probably a good idea to start with one repo initially)

```bash
if [ [ "$BUILDKITE_PULL_REQUEST" == "false" ] && [ "$BUILDKITE_PULL_REQUEST_REPO" != "git://github.com/dlang/ci.git" ] ] ; then
   wget https://github.com/dlang/ci/raw/master/buildkite.sh
fi

bash buildkite.sh | buildkite-agent pipeline upload
```

- [x] setup more agents (it seems like some projects do need quite a lot of memory, so we probably can't run too many agents on a single machine)

```bash
sudo apt-get update
sudo apt-get install -y apt-transport-https sudo
sudo sh -c 'echo deb https://apt.buildkite.com/buildkite-agent stable main > /etc/apt/sources.list.d/buildkite-agent.list'
sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 32A37959C2FA5C3C99EFBC32A79206696452D198
sudo apt-get update && sudo apt-get install -y buildkite-agent
# add token (see buildkite)
sudo systemctl enable buildkite-agent@1 && systemctl start buildkite-agent@1
sudo systemctl enable buildkite-agent@2 && systemctl start buildkite-agent@2
# ...
# install global dependencies for the build
sudo apt-get install python python-yaml make gdb g++ gcc libcurl3-dev libssl1.0-dev
```

- [x] copy all project configuration
- [x] make sure that all projects pass (currently the errors only seem to be due to missing memory)
- [ ] figure out how to make the build logs publicly accessible
- [ ] figure out why the CI notifications aren't displayed here

Other notes:
- dub doesn't build in release mode with dmd master (to be investigated)

PR #226, opened 2018-07-08
Add lubeck to the Project Tester

wilzbach:
FYI @9il @Laeeth

### DIFF
--- a/vars/runPipeline.groovy
+++ b/vars/runPipeline.groovy
@@ -465,6 +465,7 @@ def call() { timeout(time: 1, unit: 'HOURS') {
         "dlang/undeaD",
         "economicmodeling/containers",
         "ikod/dlang-requests",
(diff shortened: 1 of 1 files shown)


9il:
What lapack & cblas library is used on the CI server?

wilzbach:
Not sure -> https://github.com/dlang/ci/blob/master/ansible/ci.yml#L122
Though I think Ubuntu is used, but I don't know which version. Is this important for you? (it might change soon: https://github.com/dlang/ci/pull/225)

9il:
Travis and probably the CI has deprecated version of ubuntu with deprecated openblas. In the same time openblas library is the CBLAS/LAPACK library for default mir-blas and mir-lapack configuraitons. Lubeck will update the dependencies soon.

But probably it will not affect the CI because it uses travis. So it should be fine if travis will be updated too.

PR #228, opened 2018-07-08
Add LDC to the project tester

wilzbach:
Not sure whether this works directly on Jenkins or requires further intervention.
Depending on the time it takes to run the testsuite, maybe it's enough to just run the fail_compilation / compilable tests.

### DIFF
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -201,6 +201,14 @@ case "$REPO_FULL_NAME" in
         # remove once https://github.com/rejectedsoftware/ddox/pull/212 is merged and released
         sed 's/bridgeip.*/exit \$failure/' -i travis-ci.sh # build without the phantomcss tester
         ;;
+
+    ldc-developers/ldc)
+        git submodule update --init
+        mkdir build && cd build
+        cmake -DCMAKE_BUILD_TYPE=Debug -DD_COMPILER="$DC" ..
+        make -j2 ldc2
+        ;;
+
     *)
     use_travis_test_script
     ;;
--- a/ansible/ci.yml
+++ b/ansible/ci.yml
@@ -140,6 +140,7 @@
         - libxml2-dev
         - libxslt1-dev
         - libzmq3-dev
+        - llvm
         - mongodb-server
         - moreutils # sponge et.al.
         - pkg-config
--- a/ansible/ci.yml
+++ b/ansible/ci.yml
@@ -119,6 +119,7 @@
       apt: { name: "{{ item }}", install_recommends: no, update_cache: yes, cache_valid_time: 3600 }
       with_items:
         - build-essential
+        - cmake
         - curl
         - gdb
         - git
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -50,6 +50,7 @@ projects=(
     "vibe-d/vibe-core+select" # 3m30s
     "higgsjs/Higgs" # 3m10s
     "rejectedsoftware/ddox" # 2m42s
+    "ldc-developers/ldc" # 2m15s
     "BlackEdder/ggplotd" # 1m56s
     "eBay/tsv-utils" # 1m41s
     "dlang-community/D-Scanner" # 1m40s
@@ -69,7 +70,6 @@ projects=(
     "dlang-community/libdparse" # 13s
     "BBasile/iz" # 12s
     "dlang-community/dfmt" # 11s
-    "ldc-developers/ldc" # TODO
     # run in under 10s sorted alphabetically
     "Abscissa/libInputVisitor"
     "ariovistus/pyd"
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -69,6 +69,7 @@ projects=(
     "dlang-community/libdparse" # 13s
     "BBasile/iz" # 12s
     "dlang-community/dfmt" # 11s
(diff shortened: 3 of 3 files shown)


MartinNowak:
Does LDC require DMD to bootstrap? I'd expect it to use an old version of LDC for bootstrapping.
Did they have issues with breaking changes/bugs in dmd upstream beforehand?

Just asking whether it's a good fit for the project tester.

wilzbach:
Apparently they did have a few issues with the new `extern(C++)` mangling changes: https://forum.dlang.org/post/kovkbpkpfikmfykgqiec@forum.dlang.org

Does LDC require DMD to bootstrap? I'd expect it to use an old version of LDC for bootstrapping.

I would assume they test self-hosting too, which is more or less what this is trying to do. If you can't build ldc with dmd anymore, chances are good that a breaking change in the frontend happened.

wilzbach:
BTW looks like cmake isn't installed, so let's better wait with this until we have fully migrated to buildkite (https://github.com/dlang/ci/pull/225)?

joakim-noah:
@kinke, want to chime in on exactly how best to set this up and what problems you had?

kinke:
The recent problem turned out to be a minor one (LDC-specific `extern(D)` dtor of `extern(C++)` class not `final` anymore). The bigger problems I had recently were the D `long` C++ mangling change on OSX and, some more months back, an issue for MSVC targets which nearly broke compatibility with all existing D compiler at the time (but I don't remember the details).

What's the plan here. just building LDC for Linux x64?

wilzbach:
> What's the plan here. just building LDC for Linux x64?

Yep for the time being. However, as mentioned above it's currently failing due to `cmake` being missing and I think it's just easier to install it once globally than always rebuilding from scratch.

FYI: Once we switch to buildkite we could easily add Windows or OSX hosts to (though there are no immediate plans for this).

kinke:
Yep, you'll need CMake and a prebuilt LLVM. You can use the distro's LLVM or [our package](https://github.com/ldc-developers/llvm/releases). Just building the LDC executable, without druntime/Phobos etc. (`make ldc2`), takes something like 2-3 minutes for CircleCI, just to give you an idea.

kinke:
The build commands should look something like this:
```
cmake -DCMAKE_BUILD_TYPE=<Release|Debug> -DLLVM_ROOT_DIR=.../llvm -DD_COMPILER=.../dmd .../ldc-src
make -j<N> ldc2
```

kinke:
It'll be the `llvm-dev` package; you might also need `zlib1g-dev` for LLVM, I don't remember. [See https://github.com/ldc-developers/ldc/blob/master/.semaphoreci/setup.sh#L12 for our Ubuntu 18.04 setup. And too bad BuildKite requires registration, for pure viewing too.]

wilzbach:
> It'll be the llvm-dev package; you might also need zlib1g-dev for LLVM,

It doesn't seem to be necessary. I guess that's because libclang-dev was already installed:

https://github.com/dlang/ci/blob/4a156a39e771bd8dfa64a7507f7d98290750abf4/ansible/ci.yml#L129

> And too bad BuildKite requires registration, for pure viewing too.

Yes, I'm aware that this really sucks :/
They plan to plan to be rolling out public log file viewing soon though...
(and if you send me a ping with your email, I'm more than happy to invite you to DLang Team on Buildkite if you want to join.)

Anyhow it's passing already:

image
...

And I'm even seeing a few deprecations here:

image

That's great from the Project Tester PoV because it means that DMD's dev team can't merge PRs that breaks stuff at full pace, but needs to wait for the full deprecation cycle (which is now even more enforced!).

> Just asking whether it's a good fit for the project tester.

The debug build takes about 2m15s. I'm not sure whether this will be super helpful as most of LDC's D code should have already been tested extensively (i.e. the D Frontend part) and I doubt that the LDC team will want to use DMD for boostrapping, but there are a few customized bits and as @kinke explained there were a few mangling errors a while ago.
Also as we want to get D to be more mature, the compilers should be capable to build itself (even if it's just for the point of being able to build it.)

joakim-noah:
The idea is that the evolving C++ support won't break anything in the LDC to llvm interface.

wilzbach:
@joakim-noah @kinke I created a dummy account with readonly permissions that can be used while the logs aren't public:

user: dummy@dlang.io
password: dlangrocks

(one can't do much with this account, except for being a troll and changing the password. The only effect would be that dummy account can no longer be used.)

buildkite/build_project.sh
```diff
@@ -201,6 +201,14 @@ case "$REPO_FULL_NAME" in
         # remove once https://github.com/rejectedsoftware/ddox/pull/212 is merged and released
         sed 's/bridgeip.*/exit \$failure/' -i travis-ci.sh # build without the phantomcss tester
         ;;
+
+    ldc-developers/ldc)
+        git submodule update --init
```
joakim-noah: Would be nice to not download the stdlib unnecessarily every time: does replacing this with `rm runtime/CMakeLists.txt` work instead?

kinke: Nope. I tried it myself after seeing this PR (plain ldc build without submodules), almost got it to work eventually but then dismissed the idea (Idon't recall the actual reason anymore though, but it didn't seem worth it).

wilzbach:
Let's give this a try (:

PR #230, opened 2018-07-09
Add dxml to the project tester

wilzbach:
In theory that should be all that it takes. I might need to adjust the position based on how long the testsuite takes to run.

CC @jmdavis

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -65,6 +65,7 @@ projects=(
     "dlang-community/DCD" # 23s
     "weka-io/mecca" # 22s
     "CyberShadow/ae" # 22s
+    "jmdavis/dxml" # 22s
     "libmir/mir-algorithm" # 17s
     "dlang-community/D-YAML" # 15s
     "libmir/mir-random" # 13s


jmdavis:
Well, this looks fine, but since per your comments in the newsgroup, I need to add a .travis.yml file, this won't do much yet.

wilzbach:
Ah no need. We can directly add custom commands here too (I pushed an update).
Sorry, I blindly assumed that your project already had one, because almost everyone seems to be using Travis these days.

jmdavis:
Well, I clearly got the configuration wrong somehow, because it's running `dub test --compiler=dmd` when I configured it to do

    script:
        - dub test --build=doTests --compiler=${DC}
        - dub test --build=doReleaseTests --compiler=${DC}

Maybe you should just switch it back to whatever it was you did locally to make it run `./testRelease.sh`, and I can figure out what on earth to do with the travis file later.

wilzbach:
Ah your `.travis.yml`  is absolutely correct. It's just that the project tester is checking out the latest git tag and so far the .travis.yml is only on master. Thus it can't find it and falls back to `dub test`.
If you plan to do a release in the next few days anyhow, you could just give me another ping here and then it should automagically be green once restarted (or if you don't plan to so I can still revert to the manual lines). Let me know what you prefer.

FYI: I just verified this against the [actual extraction script](https://github.com/wilzbach/ci/blob/e9e2995c45b48a0c76ed17129772e9730b2da51c/buildkite/travis_get_script) used.

jmdavis:
Ah, so it's going off of the latest release and not master? I thought that you'd said that it went off of master. Well, going off of the latest release is probably better anyway, since it's what most folks should be using, and it means that I'm not going to accidentally screw up the tester here with commits that aren't in a release yet.

I've been in the process of preparing for another release and am expecting to get one out within the next few days.

wilzbach:
> Ah, so it's going off of the latest release and not master? 

Yes latest release. Master is only used if no tags are used at all.

> and it means that I'm not going to accidentally screw up the tester here with commits that aren't in a release yet.

That's the idea ;-)

I just rebased this as we switched from Jenkins to Buildkite recently. Unfortunately the logs are still private, which is why I sent out an invitation to you.

wilzbach:
@jmdavis FYI: dxml should now be protected against regressions ;-)

PR #232, opened 2018-07-11
Revert "Purge all dub packages"

wilzbach:
Reverts dlang/ci#231 (after the purges have been done)


wilzbach:
Jenkins is no longer in use.

PR #233, opened 2018-07-11
Fix buildkite distribution bootstrapping

wilzbach:


wilzbach:
Merging this as this blocks testing with dmd.

PR #244, opened 2018-07-16
Rename tsv-utils-dlang -> tsv-utils

wilzbach:
CC @JonDegenhardt

PR #245, opened 2018-07-26
Update tsv-utils-dlang -> tsv-utils

wilzbach:
Looks like I only did the rename partially.

See also: https://github.com/dlang/ci/pull/244


wilzbach:
CC @jondegenhardt  - sorry about missing these two instances.

PR #252, opened 2018-07-29
Cleanup after building the distribution

wilzbach:


wilzbach:
It's already done for after each individual project:

https://github.com/dlang/ci/blob/868bf9ba0f62818eae42e906da5ea7cc691b723a/buildkite/build_project.sh#L208-L209

PR #257, opened 2018-07-30
Automatically retry failing builds

wilzbach:
https://buildkite.com/docs/pipelines/command-step#automatic-retry-attributes


wilzbach:
I think so too, let's give this a try!

MartinNowak:
"Try until you randomly succeed", sounds rather blunt to me, in particular for a CI.
It might hide failures that intrinsically occur at random, e.g. memory corruptions.

Better try fine-grained retries at places where failures are expected.

Failures are never spurious btw, it's just that we don't understand or haven't fixed their cause(s).

PR #259, opened 2018-07-31
Run style target on Buildkite

wilzbach:


buildkite.sh
```diff
@@ -27,7 +27,27 @@ steps:
...
+EOF
+        ;;
+    *)
+        ;;
+esac
 
+################################################################################
+
+cat << 'EOF'
   - wait
```
wilzbach: This will now also block on the `style` target. It looks like the only two ways around this are:
- new pipeline
- after this `wait`ing step

buildkite.sh
```diff
@@ -27,7 +27,27 @@ steps:
...
+
+case "${BUILDKITE_REPO:-x}" in
+    "https://github.com/dlang/dmd.git" | \
+    "https://github.com/dlang/druntime.git" | \
+    "https://github.com/dlang/phobos.git")
+
+cat << 'EOF'
+steps:
+  - command: |
+        make -f posix.mak style
```
wilzbach: Yes, this is only run for the respective repository, but there's _zero_ that a PR to DMD will change the source code of Phobos and thus break it's style target (well, it could break the build of dscanner, but that's separately tested.)

Issue #19, opened 2017-02-18
integrate coverage testing


wilzbach:
Generate Coverage report
---------------------

This should nearly at the end as it messes up the build repositories. It should be something similar to:

```
stage ('Test coverage') {
   def repo_name = /python3 -c "import sys, json; print(json.load(sys.stdin)['repository']['name']).toupper()"/
   switch (repo_name) {
      case 'dmd':
          sh 'mv src/dmd src/host_dmd'
          sh 'make -j$N -C dmd/src -f posix.mak MODEL=$MODEL HOST_DMD=./host_dmd clean'
          sh 'make -j$N -C dmd/src -f posix.mak MODEL=$MODEL HOST_DMD=./host_dmd dmd.conf'
          sh 'make -j$N -C dmd/src -f posix.mak MODEL=$MODEL HOST_DMD=./host_dmd ENABLE_COVERAGE=1'
          sh 'make -j$N -C dmd/test MODEL=$MODEL ARGS="-O -inline -release" DMD_TEST_COVERAGE=1'
          sh 'mv src/host dmd src/dmd'
          break;
      case 'druntime':
          sh 'TEST_COVERAGE="1" make -j$N -C . -f posix.mak MODEL=$MODEL unittest-debug'
          break;
      case 'phobos':
          sh 'ENABLE_COVERAGE="1" make -f posix.mak MODEL=$MODEL unittest-debug'
          // tests needs to be run separately due to https://issues.dlang.org/show_bug.cgi?id=16397
          sh 'make -f posix.mak $(find std etc -name "*.d" | sed "s/[.]d$/.test")'
          break;
      default:
   }
```

CodeCov
-------------

Assuming the CodeCov tokens (can be fetched at e.g. https://codecov.io/gh/dlang/phobos/settings) are loaded in the environment something as `CODECOV_TOKEN_PHOBOS` (and similarly for dmd and druntime) like this should work:

```
stage ("Upload coverage report") {
    def repo_name = /python3 -c "import sys, json; print(json.load(sys.stdin)['repository']['name']).toupper()"/
    sh 'curl -s https://codecov.io/bash | bash -s - -t CODECOV_TOKEN_${DMD}'
}
```

CodeCov uptime doesn't seem to be perfect, we might want to provide a fallback URL as well - see e.g.
https://github.com/dlang/phobos/pull/5140

(though I am not sure what's the proper behavior as the CodeCov API is then probably down as well)

MartinNowak:
[...]

PR #260, opened 2018-07-31
Run CodeCov on Buildkite

wilzbach:
A first step at fixing https://github.com/dlang/ci/issues/19.
The idea is to unite a few CIs and CircleCi is one of easier.

However, we probably want to define the exact Coverage steps as a `coverage` target in the individual repositories (or maybe even as a `codecov` target as the submit step is also different.) -> `codecov.sh` script

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -1,16 +1,6 @@
 #!/bin/bash
 
-cat << 'EOF'
-steps:
-  - command: |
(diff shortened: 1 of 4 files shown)


wilzbach:
DMD PR: https://github.com/dlang/dmd/pull/8543 
(Druntime and Phobos will be easier and can be done afterwards.)

PR #261, opened 2018-07-31
[WIP] Run bootstrapping tests on Buildkite

wilzbach:
Once this is working, it would allow to retire SemaphoreCi.
OTOH the four parallel builds we get at SemaphoreCi are really fast, so maybe we don't want to do this until we have more agents or flexible scaling.


buildkite.sh
```diff
@@ -21,13 +11,57 @@ steps:
...
+done
+    ;;
+    *)
+    ;;
+esac
+
+################################################################################
+
+cat << 'EOF'
   - wait
```
wilzbach: So ideally we could start these tests in the beginning (they are totally independent to the distribution step and the following projects.)
However, it looks like a wait is a global one and thus we seem to only have two options:
- add them after the `wait` on the distribution
- add the bootstrapping projects as separate pipeline

wilzbach:
Rebased and changed, s.t. for now only DMD x86_64 + LDC x86_64 bit will be tested. Also as gdb 8 is missing on the agents, a few tests fail. I removed them for now in a similar way to https://github.com/dlang/ci/pull/291 and https://github.com/dlang/ci/pull/301.
We still need to figure out a way to fetch gdc binaries which can be used locally.

PR #262, opened 2018-07-31
Use mv instead of cp when cloning the repositories

wilzbach:
Should be a bit faster and avoid any potential conflicts by having a clean root folder.


wilzbach:
Required for https://github.com/dlang/dmd/pull/8529

PR #270, opened 2018-08-07
Revert "Temporarily disable unit-threaded"

wilzbach:
Reverts dlang/ci#269 (can be enabled once the unit-threaded build errors have been resolved.)

CC @atilaneves

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -138,7 +138,7 @@ projects=(
     "rejectedsoftware/diet-ng" # 56s
     "mbierlee/poodinis" # 40s
     "dlang/tools" # 40s
-    #"atilaneves/unit-threaded" #36s
+    "atilaneves/unit-threaded" #36s
     "d-gamedev-team/gfm" # 28s
     "gecko0307/dagon" # 25s
     "dlang-community/DCD" # 23s
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -295,6 +295,11 @@ case "$REPO_FULL_NAME" in
         use_travis_test_script
         ;;
 
+    atilaneves/unit-threaded)
+        export TERM="${TERM:-xterm-256color}"
+        use_travis_test_script
+        ;;
+
     *)
         use_travis_test_script
         ;;


atilaneves:
As mentioned in the [PR that disabled unit-threaded](https://github.com/dlang/ci/pull/269) the logs that prompted it were [these ones](https://buildkite.com/dlang/ci/builds/173#cfaf1389-9f04-415f-8d70-ed5196491cdb), which show `TERM: environment variable not set`.

That has nothing to do with unit-threaded at all, TERM doesn't even get mentioned in the code, and no breaking changes have been made.

If anything this seems like a buildkite misconfiguration.

wilzbach:
Sorry for disabling the project immediately, but it's easier to debug and find the cause here and not affect other people with red CI failures.

> If anything this seems like a buildkite misconfiguration.

We actually host the agents ourselves and they are standard Debian 9 machines.

> That has nothing to do with unit-threaded at all, TERM doesn't even get mentioned in the code, and no breaking changes have been made.

Okay, when manually exporting `TERM`, this error/warning doesn't appear anymore, but the linking still fails.

> no breaking changes have been made.

I see that v0.7.47 was tagged 2 days ago, but v0.7.46 was on May 5th
However, looking at  https://github.com/atilaneves/unit-threaded/compare/v0.7.46...v0.7.47, I don't see anything suspicious
Maybe it's the missing `--compiler=$DC` in `test.sh`, but I don't think so.
I guess it's more likely that the linking fails due to it running out of memory (the newer buildkite agents have 30G of memory + 15G swap memory), so that's unlikely too :/

atilaneves:
My main point is that since unit-threaded (and automem) have been added to the list of projects that the new dmd tries to build, I've been extremely careful to not break anything. The latest tagged version of unit-threaded works. So does master.

I already have to manage Travis CI and AppVeyor (both currently green) configurations, I'm not adding buildkite to the mix. There's always a finite probability of spurious failures, and the more pipelines there are the more likely there will be fake reds. Like this one.

When and if I break Travis or AppVeyor, then I'll take a look at it. Or maybe if there's a Github issue with "doesn't link on Debian 9".

atilaneves:
I had a thought earlier: the TERM thing is probably due to unit-threaded printing colours to the terminal. Why buildkite would fail because of that I don't know.

PR #275, opened 2018-08-13
Make LDC master build itself, using ninja

joakim-noah:
@kinke noted when updating ldc that the issue was with a self-compiled ldc crashing because of some C++ interfacing issues, so I tried to recreate that as much as possible. This will make this test run more than twice as long, but I'm not sure the previous version would have caught C++ interfacing issues.

### DIFF
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -33,6 +33,10 @@ case "$REPO_URL" in
         # for https://github.com/vibe-d/vibe.d/pull/2183, required until 0.8.5 is released
         latest_tag=master
         ;;
+    https://github.com/ldc-developers/ldc)
+        # ldc doesn't really do point releases, so master is easier to adapt if needed.
+        latest_tag=master
+        ;;
     *)
         ;;
 esac
@@ -206,10 +210,16 @@ case "$REPO_FULL_NAME" in
         ;;
 
     ldc-developers/ldc)
-        git submodule update --init
+        cd runtime && git clone --depth 5 https://github.com/ldc-developers/druntime
+        git clone --depth 5 https://github.com/ldc-developers/phobos
+        cd .. && git submodule update
+        mkdir bootstrap && cd bootstrap
+        cmake -GNinja -DCMAKE_BUILD_TYPE=Debug -DD_COMPILER="$DC" ..
+        ninja -j2 ldmd2 druntime-ldc-debug phobos2-ldc-debug
+        cd ..
         mkdir build && cd build
-        cmake -DCMAKE_BUILD_TYPE=Debug -DD_COMPILER="$DC" ..
-        make -j2 ldc2
+        cmake -GNinja -DCMAKE_BUILD_TYPE=Debug -DD_COMPILER="$(pwd)/../bootstrap/bin/ldmd2" ..
+        ninja -j2 ldc2 druntime-ldc phobos2-ldc
         ;;
 
     *)
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -99,6 +99,7 @@ projects=(
     "vibe-d/vibe.d+vibe-core-examples" # 9m51s
     "vibe-d/vibe.d+libevent-tests" # 8m35s
     "vibe-d/vibe.d+vibe-core-tests" # 6m44s
+    "ldc-developers/ldc" # 4m49s
     "vibe-d/vibe.d+libevent-base" # 4m20s
     "vibe-d/vibe.d+vibe-core-base" # 4m31s
     # https://github.com/vibe-d/vibe.d/issues/2157
@@ -109,7 +110,6 @@ projects=(
     "vibe-d/vibe-core+select" # 3m30s
     "higgsjs/Higgs" # 3m10s
     "rejectedsoftware/ddox" # 2m42s
-    "ldc-developers/ldc" # 2m15s
     "BlackEdder/ggplotd" # 1m56s
     "LaurentTreguier/dls" # 1m55s
     "eBay/tsv-utils" # 1m41s


buildkite/build_project.sh
```diff
@@ -218,10 +218,14 @@ case "$REPO_FULL_NAME" in
         ;;
 
     ldc-developers/ldc)
-        git submodule update --init
+        git submodule update --init --depth 1 runtime/druntime runtime/phobos
```
joakim-noah: This should take much less bandwidth for every checkout for each CI run.

ansible/ci.yml
```diff
@@ -118,6 +118,7 @@
         - llvm
         - mongodb-server
         - moreutils # sponge et.al.
+        - ninja
```
wilzbach: FYI: I installed `ninja-build` on the respective machines

buildkite/build_project.sh
```diff
@@ -218,10 +218,14 @@ case "$REPO_FULL_NAME" in
         ;;
 
     ldc-developers/ldc)
-        git submodule update --init
+        git submodule update --init --depth 10 runtime/druntime runtime/phobos
```
wilzbach: Are we sure that `--depth 10` is always going to work and not result in:

```
fatal: reference is not a tree: 41e9dece3fc6c2bc966cfa327ae8160f67a4c569
```

As with `--depth 1`?

joakim-noah: No, I've had it fail up to 15-25 too, don't know git well enough to say why.

wilzbach: Ah then maybe just remove `--depth` for now and we can figure this out in a different PR?

wilzbach: See also: https://stackoverflow.com/questions/2144406/how-to-make-shallow-git-submodules

wilzbach: Installed git versions on the clients are 2.7.4 and 2.11.0

kinke:
Heh yeah right, that was due to a breaking change of semantics (dtors of extern(C++) classes now virtual unless explicitly `final`). Stuff like that (benevolent in that particular case IMO) would really require a 2nd build with itself, but also make merging harder (to pass the tests, LDC would need to adapted first).

joakim-noah:
Thinking about this some more, this is not quite the same as your original situation, where you were building the latest ldc with the latest ldc, whereas this buildkite CI builds the last ldc tag, 1.10 right now, with the latest dmd. So the self-bootstrap here uses the old `extern(C++)` support from 1.10 when ldc builds itself, whereas your problem was you were using the new `extern(C++)` support in 1.11 when ldc built itself. However, it's possible that the new C++ support in dmd master will affect the ldc codegen in such a way that it affects ldc 1.10 building itself, even though ldc 1.10 uses the old C++ support.

In that case, this ldc 1.10 bootstrap simply becomes a stress test to make sure everything's working, arguably excessive beyond just building the ldc 1.10 stdlib and tests. However, it also presents no merging issues for dmd commits and won't require adapting ldc, as dmd master should always produce a working ldc 1.10. What the dmd testsuite covers is irrelevant, as the entire point of putting ldc on this buildkite CI is to potentially catch incompatibilities that may have slipped through the dmd testsuite.

I think this stress test is worth it to make sure of that, but I can scale it back if you guys think otherwise. Let me know.

I don't know what's up with the CI not testing this pull anymore.

wilzbach:
> I don't know what's up with the CI not testing this pull anymore.

Try rebasing to master. I did a few improvements yesterday that among other things will always merge the PR with upstream/master. Unfortunately this wasn't done before.

wilzbach:
FYI: it's still failing with:

```
/var/lib/buildkite-agent/builds/buildkite-agent-02-1/dlang/ci/build/ldc-developers-ldc/runtime/druntime/src/core/stdc/stdarg.d(111): Error: the `LDC_va_start` pragma is only allowed on template declarations
```

joakim-noah:
Heh, have we found our first regression from adding LDC to the CI? ;)

kinke:
> However, it's possible that the new C++ support in dmd master will affect the ldc codegen in such a way that it affects ldc 1.10 building itself, even though ldc 1.10 uses the old C++ support. [...] In that case, this ldc 1.10 bootstrap simply becomes a stress test to make sure everything's working

Yep that's what I meant. The dtor messing up the vtable was/is a (not really needed) LDC-specific addition in the D frontend, so that compiling LDC with that dtor-semantics change would break LDC at runtime (when encountering an `extern(C++)` class with dtor), so that testing the built LDC would be necessary to detect it. Building itself again is one way of doing that (particularly useful for C++ interop), taking relatively little time, but of course by no means extensive.

> Heh, have we found our first regression from adding LDC to the CI? ;)

Nope, that is the virtual-dtor breaking change, I remember it quite well - this kind of strange errors can happen when the vtbl is messed up. ;) - LDC 1.11 (with the dtor in question now being `final`) will be required, illustrating the problem I mentioned.

joakim-noah:
OK, I see what you mean about ldc changing now: sometimes changes in dmd master require modifying the buildkite projects and not just delaying the dmd pull for later, something ldc doesn't really do since it almost never puts out point releases. I just tested building the latest stable release of ldc, 1.10, with the last point release of dmd, 2.081.1, on linux/x64 and reproduced this error, which shows we don't really care about building ldc with dmd.

However, this buildkite CI error does exhibit how we can flag such regressions early, so you don't have to go digging through dmd commits to figure out what happened when 2.081 is released, but will be notified early when that dmd change was made. We can then add a `dmd-1.10` branch of ldc, commit the necessary change to get ldc building with dmd master again, and tag it as 1.10.1, so that the buildkite CI picks it up, even if we never do a binary release for that git tag. If it happens a lot, maybe we could switch this CI to use ldc master.

I suggest we hold off on this pull until the upcoming 1.11 release and if it passes, merge this pull then, to give both the dmd and ldc teams greater visibility on these errors much earlier in the development process.

wilzbach:
> If it happens a lot, maybe we could switch this CI to use ldc master.

Using `master` would be no problem if that's easier for you.
It simply requires adding the LDC github repo to this `master` checkout list:

https://github.com/dlang/ci/blob/4f7c7818bdaf81ad4a708a9cd6440a1245fc06b9/buildkite/build_project.sh#L32-L34

We typically try to avoid checking out `master` as we don't want to do alpha testing for other projects, but given that LDC has quite a few CIs in place I doubt that there will be many scenarios where a new commit to LDC will break self-bootstrapping.

kinke:
> However, this buildkite CI error does exhibit how we can flag such regressions early, so you don't have to go digging through dmd commits to figure out what happened when 2.081 is released, but will be notified early when that dmd change was made.

Yeah, that'd be nice indeed. Other common regressions, non-adapted C++ headers after modifying a D file, sadly won't be caught this way.

Building LDC master would probably be the most feasible option.

> given that LDC has quite a few CIs in place I doubt that there will be many scenarios where a new commit to LDC will break self-bootstrapping

Absolutely. Self-bootstrapping is CI-tested for Linux (Ubuntu 14.04), OSX and Windows, so that should be pretty robust.

joakim-noah:
OK, updated this pull to build master, let's see what the CI says.

joakim-noah:
Alright, finally got ldc master to build itself and added the doubled build time to `buildkite.sh`, pull's ready to go.

buildkite/build_project.sh
```diff
@@ -206,10 +210,15 @@ case "$REPO_FULL_NAME" in
         ;;
 
     ldc-developers/ldc)
-        git submodule update --init
+        cd runtime && git clone --depth 1 https://github.com/ldc-developers/druntime
+        git clone --depth 1 https://github.com/ldc-developers/phobos
+        cd .. && mkdir bootstrap && cd bootstrap
```
kinke: Wait, so you're using the latest `ldc` heads for the 2 submodules? That's not really master and might be problematic at some point (because untested in this particular combination).

kinke:
> usually

That's the point, not always, and we know how unnerving unrelated CI breakages are.

joakim-noah:
I thought we were going to keep those submodule hashes updated from now on? ;) As for using `git submodule`, see third hidden comment above and the SO link: for some reason, it usually doesn't work for me with Phobos.

kinke:
Heh yeah hopefully we are - but still, there's the chance they are out of sync for some time (serial pushes to the different repos, while an ldc commit itself is atomic, e.g., submodules being updated while the main ldc repo was being cloned).

Downloading a zipped archive from GitHub by commit hash could work, e.g., `https://github.com/dlang/phobos/archive/ded9b79.zip` (that's `HEAD~2`), if you really care about bandwidth.

buildkite/build_project.sh
```diff
@@ -206,10 +210,16 @@ case "$REPO_FULL_NAME" in
         ;;
 
     ldc-developers/ldc)
-        git submodule update --init
+        cd runtime && git clone --depth 5 https://github.com/ldc-developers/druntime
+        git clone --depth 5 https://github.com/ldc-developers/phobos
+        cd .. && git submodule update
```
joakim-noah: Alright, this should address your concerns.

joakim-noah: Unrelated CI failure, LDC builds fine.

wilzbach: The failure was due to regression that managed to sneak into Phobos - see e.g. https://github.com/dlang/ci/pull/286

I restarted the build.

joakim-noah:
Btw, [this company provides a free build node with unlimited builds for OSS projects (cannot be done concurrently), including one running Windows](https://www.shippable.com/pricing.html). Don't know if it can tie into Buildkite, but @kinke set up their linux/AArch64 node for LDC and it's working well.

joakim-noah:
ping, should be ready to go, especially since ldc master is primed for the 1.11 tag.

buildkite/build_project.sh
```diff
@@ -206,10 +210,16 @@ case "$REPO_FULL_NAME" in
         ;;
 
     ldc-developers/ldc)
-        git submodule update --init
+        cd runtime && git clone --depth 5 https://github.com/ldc-developers/druntime
+        git clone --depth 5 https://github.com/ldc-developers/phobos
```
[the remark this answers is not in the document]
joakim-noah: >If you push more than five commits to the LDC's druntime or Phobos fork, this will fail.

No idea what this means, they have a lot more than five commits. What this does is simply pull the last five commits, has nothing to do with pushing.

>AFAICT it's possible to tell git to do shallow clones

Read you own SO link from before, which I saw before you pointed me to it: it doesn't work that well. It kinda, sorta works if you're only checking out HEAD or a tag or something- I lost track- but people finally resort to what I did here, cloning each submodule you want separately and manually.

>waste those 2 seconds

Time is not the issue, it's bandwidth, as I've said before in this thread. Shallow cloning saves something like 97% of the bandwidth needed for the submodules, which adds up across thousands of CI runs, which is presumably why you shallow clone all these other CI repos too.

If this setup ends up being flaky, we can always stabilize it later.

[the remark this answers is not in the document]
wilzbach: > No idea what this means, they have a lot more than five commits. What this does is simply pull the last five commits, has nothing to do with pushing.

But I think it could happen that LDC master uses different checked out sub repositories?
If not, that's great, but then `--depth 1` would be even better.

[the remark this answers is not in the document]
joakim-noah: No, the ldc forks of druntime and phobos each have an `ldc` branch, which is the only branch from which ldc `master` checks out, and is the main branch that will be checked out for each submodule here. `--depth 1` may not work if someone commits to one of the submodule `ldc` branches and ldc `master` tries to check out an older submodule commit before the submodule hash is updated, so 5 tries to account for that. See above discussion.

joakim-noah:
You merge a new frontend once every two months, are we really going to worry about this?! 99.999% of the time, probably even for merges, it is within 5 commits of ldc HEAD. 

>the real gain and savings can be made if we start to cache the git repositories on the build agents

That would be best, of course.

wilzbach:
> You merge a new frontend once every two months, are we really going to worry about this?! 99.999% of the time, probably even for merges, it is within 5 commits of ldc HEAD.

Huh? Aren't the frontend merge PRs open for one to two weeks?
(E.g. https://github.com/ldc-developers/ldc/pull/2665, https://github.com/ldc-developers/ldc/pull/2752 and now https://github.com/ldc-developers/ldc/pull/2818)

kinke:
To me, everything making a CI build flaky is pure hatred, so if we know this can happen in advance, especially with the many DMD CI builds, it should be definitely avoided IMO, even if the chance is limited to a few seconds every 2 months.

wilzbach:
> Btw, this company provides a free build node with unlimited builds for OSS projects (cannot be done concurrently), including one running Windows. Don't know if it can tie into Buildkite, 

AFAICT they provide just a CI infrastructure like Travis, CircleCi etc.
Buildkite allows us to connect arbitrary machines to our own "CI cloud". So as long as they don't provide us with a machine that's running 24/7 and we can ssh into, I don't think we can tie it into Buildkite.

OTOH as we're reaching the limits of most free or open source tiers, Buildkite gives us an inexpensive option to increase the parallelism as just paying for a few boxes is a lot cheaper than the tiers of the CI companies (for example, 6 of the 10 agents run on very two cheap VPS boxes for which I paid 39$/year each.).
Also, @kinke et.al. it allows to share resources with the LDC folks (like the auto-tester does with GDC) and, of course, you're invited to use the "D build cloud" too and we plan on adding more agents soon.

joakim-noah:
>Huh? Aren't the frontend merge PRs open for one to two weeks?
(E.g. ldc-developers/ldc#2665, ldc-developers/ldc#2752 and now ldc-developers/ldc#2818)

And your point is? None of that time was on the `master` or submodule `ldc` branches or the ldc CI would be constantly broken.

>even if the chance is limited to a few seconds every 2 months.

That is flat out nuts, and we don't even know if the merges will cost those few secs.

kinke:
> That is flat out nuts

I'm not offended. I lose way too much time checking faulty CI runs, and every little counts. If you're so keen on saving bandwidth, then do it properly by downloading the archive as suggested earlier.

joakim-noah:
>I'm not offended.

Good, because that wasn't my intent. It's just an impossible standard, particularly given the CI is down much more already.

>I lose way too much time checking faulty CI runs, and every little counts.

Your time is worth a lot more than bandwidth or CI time, no doubt, but you're not running this CI so it will cost you zero. ;) For Seb or whoever is responsible, I'd like to see that suggested flakiness demonstrated, which I don't think it will.

>If you're so keen on saving bandwidth, then do it properly by downloading the archive as suggested earlier.

That seems like such a kludge, especially for such negligible concerns with the current approach.

wilzbach:
Buildkite jobs automatically restart themselves up to three times if they fail initially which does reduce a lot of flakiness and if it's just a few minutes we're talking about here, I would be fine, but AFAICT and as mentioned earlier it would more be like two weeks for the frontend upgrades, no?

joakim-noah:
> it would more be like two weeks for the frontend upgrades, no?

No, check the branches used in those pulls you linked. Do you think we keep master broken for two weeks? :)

wilzbach:
> Ah that's alright then. It should at most be 1 minute every 2 months, as we use a temporary ldc-merge-2.0xy branch for each submodule until it gets merged into the main ldc branch shortly after merging the LDC repo branch.

Ah fair enough. I wasn't aware of this, but then it could be `--depth 1`, no?

kinke:
There are some PRs with a couple of druntime/Phobos advancements, so to account for those, 5 seems reasonable.

wilzbach:
Ah okay, now I fully understand. However, if that's the case I would feel more comfortable with `--depth 50` as we really don't want things to fail spuriously.

joakim-noah:
No, 50 is ridiculous given the numbers I gave you above. If we find it to be a problem, we can always bump it up later.

wilzbach:
>  If we find it to be a problem, we can always bump it up later.

Well, that's always going to be harder to debug and find out about.
Just look at this recent LDC failure that suddenly appeared on a few Buildkite builds:

https://buildkite.com/dlang/dmd/builds/667#beb00ca1-866c-44e0-ae1e-656e064a0a2c

And now imagine that we have about 50 projects in the Project Tester (+ a few of our own targets).
If I only have to dig into every project every two months, that's still about a project failure to look at every day and thus I'm becoming very conservative with "we can fix it later". Sorry.

wilzbach:
> Just look at this recent LDC failure that suddenly appeared on a few Buildkite builds:

FYI: Temporarily disabled LDC and opened a revert PR, s.t. we can dig into the failure without affecting the CI infrastructure: https://github.com/dlang/ci/pull/299

joakim-noah:
>Well, that's always going to be harder to debug and find out about.

No, it isn't: you just bump it up to 10, then 20, etc. till it's fixed.

>If I only have to dig into every project every two months

As explained before, you won't have to dig into anything every two months.

You seem to have problems reading what is written, so I'm done repeatedly explaining all this to you.

Pull or don't pull, I don't care at this point. I'm not responding anymore, as I just don't want to deal with these nonsense objections anymore.

ZombineDev:
> You seem to have problems reading what is written, so I'm done repeatedly explaining all this to you.
Pull or don't pull, I don't care at this point. I'm not responding anymore, as I just don't want to deal with these nonsense objections anymore.

@joakim-noah There's no need to get confrontational. Everyone involved here sees clearly the benefit of doing more LDC-related testing and we all want to help on making it happen as it benefits everyone: LDC  devs won't have to deal with certain type regressions after the fact and dmd/druntime/phobos devs will have more assurance that PRs won't cause unexpected breaking changes. (Of course no finite amount of testing can prove the absence of bugs, but it *does* help.)

The reason @wilzbach seems to act more conservatively than necessary from your PoV is that during the last year there have been a number of breakages to the project tester and when they're not fixed in short time frame, developers start doubting the project tester's authority and begin ignoring it, and then regressions start sneaking it making it even harder to restore the project tester to a green state. The probability of a random project failing is still super high (just look at the build history of the past several months) and we can't afford to not taking steps to reduce this probability.

We're all contributing in our free time and given that @wilzbach does the majority of the work by himself during the past several months to keep the project tester green I think it makes sense to respect his wish to take extra steps to ensure the system runs hands off. (Of course not to underestimate Martin who put the foundations in place and other part-time contributors like yourself.)
TBH, your reaction seems a bit disproportional, given that all you were asked for was to change a single parameter and that you don't have any stakes for reducing the network traffic.

ZombineDev:
> You guys are acting as though git issues will take a bunch of debugging. They won't, they're trivial. You simply see that a checkout didn't work, you reproduce it yourself with a shallow clone, then bump the number of commits till it works. That's it.

It's not that it may be something hard to debug for people aware of the CI infrastructure, but that issues may appear when builds are triggered from a totally unrelated pull requests to other repos. We want to minimize the time some (potentially first-time) contributor's PR fails for no apparent reason and blocks their work. Even if you assume that most dmd/druntime/phobos contributors know all about the repos they're contributing to (in practice I think most people are oblivious to how the build or CI systems are set up), you can't expect them to know how each project (of the the growing BuildKit project list) is setup, and even less to go ahead and debug issue when no one is around to fix them at the moment. And even if they do, their contributing experience becomes something like this: https://www.youtube.com/watch?v=AbSehcT19u0 :smile: 

> If ldc causes any problems, put it on me to fix them.

I don't doubt that you'll be responsive and help fix any issues, it's just that:
A) BuildKite build logs are still not public, so many contributors won't even know what's broken / why the build fails.
B) Even if they do find out that  LDC is failing how would they no to contact you, any other member of the LDC team, or people with the knowledge to fix the issue in this repo.
C) What if the people with the knowledge the fix this go on a well-deserved one week vacation or simply busy with life?

I still think that the discussion here is getting blown out of proportion and I'm fine with merging the PR as it is. I'm just pointing out that @wilzbach has the full right to be conservative and hard to convince, as a higher priority for him is making sure that the system works as much as possible without his (or anyone's) intervention.

joakim-noah:
All this is answered above: it's not going to break. I've given you the reasons why, simply asserting that it might is nonsense. On the 0.001% chance it does, it's an easy fix.

And all this hand-wringing is frankly ridiculous considering that the Buildkite logs aren't available in the first place. That's a much bigger usability issue for contributors to be dealt with than this (I know you can only wait for Buildkite, but it is still much bigger).

joakim-noah:
Ping, there is no reason to delay this pull because of unreasonable concerns about the LDC stdlib falling out of sync. The LDC stdlib receives very few pulls, maybe 5-6 per month, as most work is directed upstream. Concern that it will break this CI is misplaced.

joakim-noah:
Since this pull is being ignored, closing.

PR #281, opened 2018-08-14
Load distribution for the coverage target

wilzbach:


wilzbach:
Merging directly as this can't be tested here anyhow.

PR #286, opened 2018-08-15
Revert "Temporarily disable dxml"

wilzbach:
Reverts dlang/ci#285

See e.g. https://buildkite.com/dlang/phobos/builds/180#3399564a-7cc7-4850-955c-101f35d33b6b

It looks like the skipOver PR (dlang/phobos#6143) introduced this regression.

CC @jmdavis

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -124,7 +124,7 @@ projects=(
     "dlang-community/DCD" # 23s
     "weka-io/mecca" # 22s
     "CyberShadow/ae" # 22s
-    #"jmdavis/dxml" # 22s
+    "jmdavis/dxml" # 22s
     "libmir/mir-algorithm" # 17s
     "dlang-community/D-YAML" # 15s
     "libmir/mir-random" # 13s


jmdavis:
Okay. I'm totally confused. Why would you create a PR to disable the dxml tests and then immediately create one to re-enable them? I get that a regression was introduced into Phobos that was missed, but why create a PR and then immediately create one to revert it? Is this just so that it a PR is ready once skipOver has been fixed?

wilzbach:
> Okay. I'm totally confused.

Sorry.

>  Why would you create a PR to disable the dxml tests and then immediately create one to re-enable them?

So that the CI isn't red on all systems and people loose trust in the CIs.
Often the failure/regression is a bit harder to fix (e.g. https://github.com/dlang/ci/pull/270) and keeping the CI red for a few days is imho not a good option. With this approach re-enabling the project isn't forgotten and clicking on the "Revert" button is easy.

> I get that a regression was introduced into Phobos that was missed, but why create a PR and then immediately create one to revert it? Is this just so that it a PR is ready once skipOver has been fixed?

Yes and that I/we don't forget to re-enable dxml once the regression has been fixed.

BTW PR to fix the regression: https://github.com/dlang/phobos/pull/6667

wilzbach:
Update: once https://github.com/dlang/phobos/pull/6675 has been merged (and the regression fix is in master too), this PR can be restarted and merged.

ZombineDev:
@wilzbach I restarted the build as I merged dlang/phobos#6675. You can auto-merge this PR as well.

wilzbach:
Thanks!

PR #287, opened 2018-08-16
Run Phobos's testsuite on the Project Tester

wilzbach:
Motivation: https://github.com/dlang/phobos/pull/6645#issuecomment-413491253

tl;dr: there are some things which are really hard to add to the `auto-tester`, but should really run on all dmd,druntime and phobos repositories. Adding such tests separately to the existing CircleCi setups is painful (and we want to retire CircleCi anyhow), so I think the easiest for now is just to run Phobos's testsuite here too.

For now, this PR will use the `auto-tester-test` target, but https://github.com/dlang/phobos/pull/6669 has been merged this can be switched to `buildkite-test`.


wilzbach:
This is going to be fun:

```
std.socket.SocketParameterException@std/socket.d(1986): Path too long
----------------
??:? pure @safe bool std.exception.enforce!(bool).enforce(bool, lazy object.Throwable) [0xe424be3b]
??:? pure @trusted std.socket.UnixAddress std.socket.UnixAddress.__ctor(scope const(char)[]) [0xe473d1dd]
??:? @safe void std.socket.__unittest_L2024_C11() [0xe473d54c]
??:? void std.socket.__modtest() [0xe47455fe]
```

It's probably due to us setting `TMP` to the current working directory as we run into concurrency issues when using the global `/tmp` :/

PR #288, opened 2018-08-20
Add optional to the project tester

wilzbach:
Small library, but does use a few interesting D features.


wilzbach:
FYI @aliak00. We're in the process of making the D ecosystem stabler and thus adding more projects to the Project Tester.
The testsuite of latest stable version (i.e. the latest tagged version which isn't `-alpha`) of your lib will be run on every PR and as long as you don't release a new version that doesn't pass on your CIs, it should require no interaction.

aliak00:
Awesome! Good to hear :)

PR #289, opened 2018-08-20
Add sumtype to the project tester

wilzbach:
A small lib, but it uses a few interesting D features.


wilzbach:
FYI @pbackus. We're in the process of making the D ecosystem stabler and thus adding more projects to the Project Tester.
The testsuite of latest stable version (i.e. the latest tagged version which isn't `-alpha`) of your lib will be run on every PR and as long as you don't release a new version that doesn't pass on your CIs, it should require no interaction.

PR #293, opened 2018-08-22
Fix Phobos target

wilzbach:
The failure was due to clone_repositories trying to move the existing files into the BUILDKITE_REPO folder.

This, however, doesn't work for the dlang/phobos target as it doesn't checkout phobos as in `build_project` it's unknown whether the master or stable branch or another one should be used.


wilzbach:
Merging as dlang/phobos is green and the actual problem can only be replicated for the dmd/druntime/phobos repositories.

PR #296, opened 2018-08-23
run more buildkite agents on ci.dlang.io server

MartinNowak:


wilzbach:
FYI: I will just disable LDC for now and then open a revert PR, s.t. we can have more time to figure this failure out, don't need to revert to hacks and don't create red status checks for our contributors.

https://github.com/dlang/ci/pull/298

PR #300, opened 2018-08-23
fix ldc testing

MartinNowak:
- requires llvm-dev (not just llvm)

### DIFF
--- a/ansible/ci.yml
+++ b/ansible/ci.yml
@@ -101,7 +101,7 @@
         - libxml2-dev
         - libxslt1-dev
         - libzmq3-dev
-        - llvm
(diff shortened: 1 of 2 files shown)


ZombineDev:
Looks like restarting Higgs manually (4th retry, the first 3 runs were automatic) was all that was needed to make it pass. The ability to restart a single project test is super useful! It seems that the machine on which Higgs was run needs a little bit more resources to ensure smooth operation.

wilzbach:
> It seems that the machine on which Higgs was run needs a little bit more resources to ensure smooth operation.

We can simply restrict on which machines it runs:

https://github.com/dlang/ci/pull/302

PR #301, opened 2018-08-24
Run Druntime's testsuite on Buildkite

wilzbach:
Split-off from https://github.com/dlang/ci/pull/291

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -118,6 +118,7 @@ projects=(
     "LaurentTreguier/dls" # 1m55s
     "eBay/tsv-utils" # 1m41s
     "dlang-community/D-Scanner" # 1m40s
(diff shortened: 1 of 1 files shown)


wilzbach:
This is blocked on druntime's testsuite requiring gdb 8.
`rt_trap_exceptions_drt_gdb` in particular.

wilzbach:
This is finally passing.
CC @ZombineDev - okay with patching out the test from druntime which requires gdb8 for now?

ZombineDev:
Sure, just open an issue to add GDB 8, so we don't forget to reenable the test when possible.

wilzbach:
> Sure, just open an issue to add GDB 8, so we don't forget to reenable the test when possible.

Done -> https://github.com/dlang/ci/issues/339

PR #304, opened 2018-09-03
remove jenkins runners

MartinNowak:
- keep jenkins running for now to serve historical PR results


BBasile:
Why has this been deactivated ?

wilzbach:
We have switched to buildkite about two months ago as Jenkins build were too unreliable and we had many problems with the Jenkins test runner.

BBasile:
Can i see the logs ? Maybe someone need to invite me to buildkite org for dlang (b2 DOT temp AT gmx DOT com, replace all caps by the char that's named)

PR #306, opened 2018-09-09
Add `pip3 install meson` to buildkite/setup_linux.sh

n8sh:
For testing https://github.com/libmir/mir-random.


BBasile:
```

Performing "debug" build using /var/lib/buildkite-agent/builds/buildkite-agent-04-1/dlang/ci/build/distribution/bin/dmd for x86_64.
--
  | flex_plots ~master: building configuration "library"...
  | bash: line 3: meson: command not found
  | 🚨 Error: The command exited with status 127


```

looks like you need to deactivate mir/random first, handle this PR and then reactivate.

Geod24:
Do you want to pursue this ? I think we should aim to run the `install` step in its entirety if we go this way.

wilzbach:
FYI: the Shell script is never executed and in fact was more or less accidentally committed to the repository when we changed to Buildkite. The Ansible scripts are used for setup, but for better performance all packages are preinstalled (no docker insanity) and the build just runs in a locked userspace.

PR #311, opened 2018-09-14
on-demand buildkite agents

MartinNowak:


wilzbach:
I'm more than happy to remove the VPS machines from the agents list (: 
Though I am not sure whether it might not be easier to use e.g. the OpenCollective money to fund an extra server than trying to do more complicated on-demand bootstrapping?

MartinNowak:
> Though I am not sure whether it might not be easier to use e.g. the OpenCollective money to fund an extra server than trying to do more complicated on-demand bootstrapping?

It doesn't make sense to spend 5-10x more than necessary.

PR #323, opened 2018-10-19
Fix #308 - wrong version of Phobos is checked out during testing it as a project

### LINKED ISSUE #308: Phobos PRs HEAD are tested, not their merge commit
See comment [here](https://github.com/dlang/phobos/pull/6705#issuecomment-420156579).

Phobos broke because a PR was merged when it was green, but the resulting master build was red.
That PR (incorrectly) changed the Makefile target `buildkite-test` to use `betterC` instead of `betterc` (which had changed on master but not on stable).

wilzbach:
The problem was that when Phobos (or dub or tools for that matter) is tested as a project, it used the same logic as all the other to-be-tested projects (i.e. fetching and checking out the latest tag).

With this PR the core repositories will use the same logic as its existing in the build_distribution step:
- merge the PR head into the target branch (if it's the same repo)
- otherwise check if the branch from the repo that triggered the PR exists in the to-be-tested repo (master and stable always exist)
- otherwise checkout master

### DIFF
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -41,7 +41,7 @@ case "$REPO_URL" in
         ;;
 esac
 
-# Don't checkout a tagged version of the core repostories like Phobos
+# Don't checkout a tagged version of the core repositories like Phobos
 case "$REPO_FULL_NAME" in
     dlang/phobos)
         ;;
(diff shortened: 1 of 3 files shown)


wilzbach:
Merging as this can only be tested for real on the Phobos side and the logs look good.

PR #324, opened 2018-10-22
gen image for buildkite based (nightly) release-builder

MartinNowak:
- builds releases for all OSes using prepared virtualbox images
- uploads tarballs to S3
- updates downloads.dlang.org/nightlies index pages

Also see http://downloads.dlang.org/nightlies/ and https://buildkite.com/dlang/build-release
Starting and stopping of agents will be done by dlang-bot at some point, but is still manual atm.


buildkite/build_release.sh
```diff
@@ -0,0 +1,76 @@
...
+LATEST=$(curl -fsSL http://downloads.dlang.org/releases/LATEST)
+rdmd build_all.d v$LATEST $BRANCH && success=yes || success=no
+rm -f build/*.zip # only keep lzma archives
+
+for file in build/*; do
+    gpg --detach-sign --no-use-agent $file
+done
+
+mv build dmd-$BRANCH-$TODAY
+aws --profile ddo s3 sync dmd-$BRANCH-$TODAY/ s3://downloads.dlang.org/nightlies/dmd-$BRANCH-$TODAY/ --acl public-read --cache-control max-age=604800
```
wilzbach: Should we really upload this automated into the same bucket? I mean the CI is a rather easy target vector and quite exposed.
Ideas:
1) use a different bucket or hoster (there are many free ones)
2) restrict the AWS token to the nightly folders only

MartinNowak: > 1. use a different bucket or hoster (there are many free ones)

Mmh, it makes a lot of sense to put it on the same download page as the other releases, see http://downloads.dlang.org/. It wasn't done originally as I didn't have access to ddo.

> 2. restrict the AWS token to the nightly folders only

That would be good, I'm already only synching the nightlies folder. Will ask Brad for another token.
Let's do this in a second iteration to not stall this. This is not intended to run as CI, but only via scheduler. Also I restricted buildkite access for the release-builder pipeline to admins, though nothing prevents people from scheduling other builds on that queue once the agent is running.

buildkite/build_release.sh
```diff
@@ -0,0 +1,76 @@
...
+
+git reset --hard --quiet
+git clean -df --quiet
+if [ $BRANCH != master ] && [ $BRANCH != stable ] &&
+       ! git ls-remote --exit-code --heads origin $BRANCH > /dev/null; then
+    # use master as fallback base for feature branches
+    git fetch origin master --quiet
+else
+    git fetch origin $BRANCH --quiet
+fi
```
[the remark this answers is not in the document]
MartinNowak: It's not the buildkite branch of the ci repo, but the to-be build branch.
Guess I can `BUILDKITE_BRANCH=$BRANCH ./origin_target_branch.sh` though.

buildkite/build_release.sh
wilzbach: > That would be good, I'm already only synching the nightlies folder. Will ask Brad for another token.

If the access is limited to a folder only, it should be fine. I was just worried that an accidental leak here could have an impact on all our binaries.

> Could prolly add an agent hook to verify we're on the right pipeline.
> ...
> Let's do this in a second iteration to not stall this. 

Sure. Let's just open an issue, s.t. this isn't lost?

buildkite/build_release.sh
```diff
@@ -0,0 +1,76 @@
...
+git checkout FETCH_HEAD --quiet
+
+mkdir build
+LATEST=$(curl -fsSL http://downloads.dlang.org/releases/LATEST)
```
wilzbach: Would recommend to use `wget -q -O -` as it automatically has a retry logic.

buildkite/build_release.sh
```diff
@@ -0,0 +1,76 @@
...
+rdmd build_all.d v$LATEST $BRANCH && success=yes || success=no
```
wilzbach: So the idea is to upload everything that got build as success=no could mean that only one Box failed?
Maybe a good idea to add this as a comment.

buildkite/build_release.sh
```diff
@@ -0,0 +1,76 @@
...
+TODAY=$(date +'%Y-%m-%d')
+: ${BRANCH:=master}
+
+if [ ! -d installer ]; then
+    git clone https://github.com/dlang/installer.git installer
+fi
+pushd installer/create_dmd_release
+
+git reset --hard --quiet
+git clean -df --quiet
```
wilzbach: You need a double `-ff` to remove cloned sub repositories (not sure whether you want to).
For the other jobs we do this to avoid any potential side effects from previous builds.

buildkite/build_release.sh
```diff
@@ -0,0 +1,76 @@
...
+    done
+    echo -n | aws --profile ddo s3 cp - s3://downloads.dlang.org/nightlies/dmd-$BRANCH --acl public-read --cache-control max-age=604800 --website-redirect http://downloads.dlang.org/nightlies/dmd-$BRANCH-$TODAY
+    echo -n $BRANCH-$TODAY | aws --profile ddo s3 cp - s3://downloads.dlang.org/nightlies/dmd-$BRANCH/LATEST --acl public-read --cache-control max-age=604800
+fi
+
+make -C src
+rm -rf ddo
+./src/build-gen-index s3_index generate
+aws --profile ddo s3 sync ./ddo/nightlies/ s3://downloads.dlang.org/nightlies/ --acl public-read --cache-control max-age=604800
+
```
wilzbach: Maybe do a bit of cleanup here? Not done by default in git repositories in git repositories on a `git clean` in the parent.

buildkite/build_release_pipeline.yml
```diff
@@ -0,0 +1,8 @@
+steps:
+  - label: ':package: Build'
+    command:
```
wilzbach: For the other projects I always added a `git clean -ffdxq .` at the top to be sure that nothing old was left over. Though of course this breaks any caching, but I would argue that knowing the script will always take the same path is worth this - especially if the task is only run once a day.

buildkite/build_release.sh
```diff
@@ -0,0 +1,76 @@
...
+    },
+    "s3bucket": "downloads.dlang.org"
+}
+EOS
+fi
+
+# create redirects from dmd-master/* to dmd-master-2018-10-12/*
+if [ $success = yes ]; then
+    for file in $(ls ../installer/create_dmd_release/dmd-$BRANCH-$TODAY); do
+        echo -n | aws --profile ddo s3 cp - s3://downloads.dlang.org/nightlies/dmd-$BRANCH/$file --acl public-read --cache-control max-age=604800 --website-redirect http://downloads.dlang.org/nightlies/dmd-$BRANCH-$TODAY/$file
```
wilzbach: Neat trick :+1:

buildkite/build_release.sh
```diff
@@ -0,0 +1,76 @@
...
+    done
+    echo -n | aws --profile ddo s3 cp - s3://downloads.dlang.org/nightlies/dmd-$BRANCH --acl public-read --cache-control max-age=604800 --website-redirect http://downloads.dlang.org/nightlies/dmd-$BRANCH-$TODAY
+    echo -n $BRANCH-$TODAY | aws --profile ddo s3 cp - s3://downloads.dlang.org/nightlies/dmd-$BRANCH/LATEST --acl public-read --cache-control max-age=604800
+fi
+
+make -C src
+rm -rf ddo
+./src/build-gen-index s3_index generate
+aws --profile ddo s3 sync ./ddo/nightlies/ s3://downloads.dlang.org/nightlies/ --acl public-read --cache-control max-age=604800
```
wilzbach: Maybe sync the files _first_ before you do the redirect?
(maybe the sync breaks and or the internet becomes slow ...)

buildkite/build_release.sh
```diff
@@ -0,0 +1,76 @@
...
+if [ ! -f config.json ]; then
+    (umask 0277 && cat > config.json) <<EOS
+{
+    "aws": {
+        "key": "$(sed -n 's|aws_access_key_id\s*=\s*||p' ~/.aws/credentials)",
+        "secret": "$(sed -n 's|aws_secret_access_key\s*=\s*||p' ~/.aws/credentials)",
+        "endpoint": "s3-$(sed -n 's|region\s*=\s||p' ~/.aws/config).amazonaws.com"
+    },
+    "s3bucket": "downloads.dlang.org"
+}
```
wilzbach: That looks overly complicated. Why not pass them in simply as environment variables?
(If an attacker gets access to the pipeline he can also `cat` the credentials file)

buildkite/build_release.sh
wilzbach: > Also I restricted buildkite access for the release-builder pipeline to admins, 

Be aware that the AWS credentials still might show up in the logs (or old logs) and currently "Everyone" has read access to the logs:

image

buildkite/build_release.sh
MartinNowak: Good point, this was used in the old setup to uploads build logs.
So far we never uploaded partially succeeding builds.

buildkite/build_release.sh
MartinNowak: See https://github.com/dlang/downloads.dlang.org/pull/5 to improve on this.

MartinNowak:
prerequisites:
https://github.com/dlang/downloads.dlang.org/pull/5
collaterals:
https://github.com/dlang/installer/pull/350

buildkite/build_release.sh
MartinNowak: We throw away the VM after the build anyhow and it seems more worthwhile to leave around build artifacts for debugging of failures or when invoking the script manually.

buildkite/build_release_pipeline.yml
MartinNowak: OK, let's try.

buildkite/build_release.sh
MartinNowak: That's just syncing the index.html files, the actual binaries are uploaded beforehand.

buildkite/build_release.sh
MartinNowak: Hardly any way to avoid this, environment variables can be printed as well.
I've added `~/.aws/credentials` parsing in https://github.com/dlang/downloads.dlang.org/pull/5 to reduce the risk of exposing them to the public logs though.

MartinNowak:
Done @wilzbach

PR #328, opened 2018-10-26
Add timeouts for Buildkite commands/steps

wilzbach:
I have seen the Phobos task run for over an hours a few times now.
This should prevent such issues, though of course the actual root cause should be found.

See also: https://buildkite.com/docs/pipelines/command-step


wilzbach:
Seems to be working:

image

PR #329, opened 2018-10-27
Only run dlang/phobos on high memory machines

wilzbach:
See also: https://github.com/dlang/ci/pull/328


wilzbach:
This means the Phobos test won't be run anymore on the cheaper VPS machines.

PR #330, opened 2018-10-30
Disable ariovistus/pyd

Geod24:
The repository switched to a docker-based testsuite:
https://github.com/ariovistus/pyd/commit/e21dde4fa0f12abe5def811eae6552e6c328cde5

CC @ariovistus

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -142,7 +142,7 @@ projects=(
     "dlang-community/dfmt" # 11s
     # run in under 10s sorted alphabetically
     "Abscissa/libInputVisitor"
-    "ariovistus/pyd"
+    #"ariovistus/pyd"
     "atilaneves/automem"
     "AuburnSounds/intel-intrinsics"
     "DerelictOrg/DerelictFT"


ariovistus:
well, this is the first I've heard of this project building pyd. I'll try to get into build kite later, but for the moment, I don't know how stuff works in this project or what's failing. Is it just using test scripts from my .travis.yml?

MartinNowak:
Thanks for getting back @ariovistus. We're running the project tester against a couple of often used dub packages to prevent regressions during development of dlang.
Your benefit in participating is a higher protection of your project against dlang regressions.

> Is it just using test scripts from my .travis.yml?

Yes, that's what it tries to do, though we have a couple of overrides for some projects.
Do you think it would be possile to continue using the non-docker based test script?
Would it make sense to keep it alongside your Travis-CI docker tests as well?

ariovistus:
I can bring back the old .travis.yml and keep it somewhere in my repo, if that would be suitable. Also, is this project just testing against dmd builds, or does it test multiple compilers?

ZombineDev:
@ariovistus buildkite tests each pull request to dmd, druntime, phobos and dub against all latest tag of many dub packages such as yours. See https://forum.dlang.org/thread/vytpearrfgxqfqugtaow@forum.dlang.org for more info.

@MartinNowak what would adding support for docker and other container runtimes entail?

PR #340, opened 2018-11-15
Fix Phobos style target builds

wilzbach:


thewilsonator:
This still fail on phobos

ZombineDev:
It looks like you have a bit more work to do on the phobos test target.

wilzbach:
Passing now (:

PR #344, opened 2018-11-28
Fix merging head logic and checkout a respective PR for the dlang projects

wilzbach:
See also: https://github.com/dlang/druntime/pull/2376#issuecomment-442453032

### DIFF
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -147,6 +147,12 @@ case "$REPO_FULL_NAME" in
         CONFIG=select ./travis-ci.sh
         ;;
 
+    dlang/ci)
+        echo "Check that the PR commit has been merged into the target branch"
+        echo "Current commit: $(git describe)"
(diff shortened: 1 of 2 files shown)


buildkite/build_project.sh
```diff
@@ -147,6 +147,12 @@ case "$REPO_FULL_NAME" in
         CONFIG=select ./travis-ci.sh
         ;;
 
+    dlang/ci)
+        echo "Check that the PR commit has been merged into the target branch"
+        echo "Current commit: $(git describe)"
+        [[ "$(git log --format=%B -n 1)" =~ Merge[[:space:]]${BUILDKITE_COMMIT:-invalid} ]]
```
wilzbach: A simple test, s.t. we don't accidentally break this in the future (and also actually can test this here).

wilzbach:
CC @thewilsonator @ZombineDev (this should be passing now)

PR #345, opened 2018-11-28
Rename LaurentTreguier/dls -> d-language-server/dls

LaurentTreguier:
After https://github.com/d-language-server/dls/pull/17, dls now doesn't have a fixed dub version in `dub.selections.json` anymore, but I'm not entirely satisfied with simply removing it, for three reasons:
- it goes against reproducible builds, which is something I prefer having (in case I need to find out about a regression somewhere or anything)
- when launching a simple build, dub rewrites `dub.selections.json` with all the required dependencies, adding the `dub` line again and automatically dirtying the file in the process
- I'm an airhead, and I might just re-add the line in the repo by accident when upgrading other dependencies

The 2 commits in this PR:
- rename the repository (as I moved it and other related repos under an organization)
- add an entry in `build_project.sh` to erase the `dub` line from `dub.selections.json` for dls, and let dub pick the appropriate version instead

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -117,7 +117,7 @@ projects=(
     "higgsjs/Higgs" # 3m10s
     "rejectedsoftware/ddox" # 2m42s
     "BlackEdder/ggplotd" # 1m56s
-    "LaurentTreguier/dls" # 1m55s
+    "d-language-server/dls" # 1m55s
     "eBay/tsv-utils" # 1m41s
     "dlang-community/D-Scanner" # 1m40s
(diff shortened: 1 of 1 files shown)


jacob-carlborg:
I think this is the wrong fix. See https://github.com/d-language-server/dls/pull/17 for more details.

wilzbach:
Yeah let's keep the first commit and follow up on https://github.com/d-language-server/dls/pull/17 (i.e. using `v1.12.1-alpha.1` in the `dub.selection.json` and releasing a new version).

PR #351, opened 2018-11-29
Move/copy hidden files of the base repository when needing to clone all core repositories

wilzbach:
The reason https://github.com/dlang/ci/pull/346 didn't work was that the `for f in ./* ; do` doesn't match hidden files and thus still didn't copy `.git` over.
An alternative would be `ls -a` or `find`.


jacob-carlborg:
I suggest using `find` instead.

jacob-carlborg:
BTW, `ls` should never be used in scripts, it's only for listing directories in a human readable format, not for programmatically doing something with the output.

MartinNowak:
Let's do custom checkout, this doesn't seem to lead anywhere.

PR #353, opened 2018-11-30
Avoid an unnecesary git clone for core repositories if clone_repositories is called later anyhow

wilzbach:


buildkite/build_project.sh
```diff
@@ -57,8 +57,18 @@ case "$REPO_FULL_NAME" in
...
             latest_tag="IS-ALREADY-CHECKED-OUT"
         else
-            # otherwise checkout the respective branch
-            latest_tag=$("$DIR/origin_target_branch.sh" "${REPO_URL}")
+            # for the main core repositories,
+            # clone_repositories.sh will clone them together
+            case "$REPO_FULL_NAME" in
+                "dlang/dmd" | \
+                "dlang/druntime" | \
+                "dlang/phobos")
```
jacob-carlborg: tools and dub as well?

wilzbach: No they are checked out as individual repos as they don't depend on the core repos - just an up to date DMD which is always built initially and already provided ty them.

wilzbach: This is just about these three repos as for them clone_repositories will be executed during "project testing"

jacob-carlborg: Ok.

Issue #355, opened 2018-11-30
Build druntime currently fails


wilzbach:
The error message is quite hidden:

```
benchmark/runbench.d(9): Error: module `stdio` is in file 'std/stdio.d' which cannot be read
--
  | import path[0] = /var/lib/buildkite-agent/builds/buildkite-agent-02-1/dlang/phobos/build/dlang-druntime/druntime/import
  | import path[1] = ../phobos
  | posix.mak:327: recipe for target 'generated/linux/release/64/benchmark' failed
  | make: *** [generated/linux/release/64/benchmark] Error 1
  | make: *** Waiting for unfinished jobs....
```

(from e.g. https://buildkite.com/dlang/phobos/builds/721#a9d4c49f-d5d9-4009-a816-e65ad4d50c84)

wilzbach:
This has been fixed now.

PR #357, opened 2018-12-01
Fix clone_repositories if a project repository triggered the build

wilzbach:
If a project triggers a PR (e.g. dlang/phobos), then it shouldn't be recloned
if dlang/phobos needs to be tested. This logic is already in `build_project`
and `clone_repositories`.

At the moment, builds are done in the subfolder `build`, but the
`clone_repositories` script assumes that we're at the git root folder.

This should fix many of the remaining Buildkite failures (maybe even #355).


wilzbach:
I don't really like this, but I will merge this for now in the interest of finally fixing the broken Buildkite pipelines.

PR #363, opened 2018-12-12
Unshallow git history for dlang-tour builds

wilzbach:
Should hopefully fix Buildkite.

### DIFF
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -222,6 +222,10 @@ case "$REPO_FULL_NAME" in
         ;;
 
     dlang-tour/core)
+        # this is required as the older git versions on the agents have troubles
+        # checking older submodules out otherwise
+        git pull --unshallow
+        (cd public/content/en && git fetch)
         git submodule update --init public/content/en
         dub test "--compiler=$DC"
         ;;


joakim-noah:
Maybe better to update the git packages?

wilzbach:
> Maybe better to update the git packages?

Ah it would then just fail again once a new commit gets added to dlang-tour/english.
However, it looks like even this doesn't solve the issue :/

joakim-noah:
>Ah it would then just fail again 

How? It works for me now with git 2.19.2, ie I don't need these changes.

wilzbach:
Well yeah I have no idea anymore what's going on at the agents and I ran out of ideas for now.

wilzbach:
> What is the git version used by the agent ? 

I don't know it either, but upgrading is a bit hard as they use a default Ubuntu image.

> Other option than updating the git version would be to clone without depth set to 1

Isn't `unshallow` going to fix this retro-actively?

ZombineDev:
Another reason to drop the shitty setup (w.r.t. build env, not infra) and use Docker ;)

joakim-noah:
#365 reports `git version 2.11.0`, [which was released more than two years ago](https://blog.github.com/2016-11-29-git-2-11-has-been-released/).

wilzbach:
No Martin just has some auto-adjusting of agents in place. Though it looks like its not working properly.

joakim-noah:
Fixed in #366, I don't know what's up with the other killed builds.

joakim-noah:
Can be closed, already fixed.

joakim-noah:
btw, ldc is now timing out on Buildkite, didn't used to before. It uses two cores to build itself, `ninja -j2`: can that be increased to make sure it doesn't hit the 40 min. timeout or will it not help on the hardware being used?

wilzbach:
I honestly don't know whether this would help, but it's worth a try. Otherwise we could increase the timeout to 60 minutes.

PR #370, opened 2018-12-22
enable dub's experimental build cache improvements

MartinNowak:
- also see https://github.com/dlang/dub/pull/1589 and
  https://github.com/dlang/dub/pull/1624
- reduces test time for vibe-d's examples by ~50%


MartinNowak:
> Shouldn't we enable this when we build dub for the release distribution archive? This is just for dub's testsuite...

Thanks for catching this @wilzbach. Quite a lot of scripts ;), should be at the right place now.

PR #371, opened 2018-12-29
Retry nightly builds up to 3 times

wilzbach:
Random failures also seem to happen here, the retry logic helped us quite a lot for the CI pipelines, so I think it could be benefitial here too.

See e.g. https://buildkite.com/dlang/build-release/builds/112#25005182-8ce8-46f7-a2c6-4ebd94b2c960

### DIFF
--- a/buildkite/build_release_pipeline.yml
+++ b/buildkite/build_release_pipeline.yml
@@ -8,3 +8,6 @@ steps:
       - './buildkite/build_release.sh'
     agents:
       queue: release-builder
+    retry:
+      automatic:
+        limit: 2


MartinNowak:
No, it doesn't make sense to retry an hour-long build with an unclear error condition. The failure today was due to the server loosing access to it's network-fs, older failures were due to VM-boot timeouts.
The former would require a server reboot, the latter should be handled in the build script.

MartinNowak:
A similar argument goes for the random retries in the project tester, that cause a 3x more workload for broken PRs. We should rather identify flaky tests and deal with them on a detailed level.

PR #373, opened 2019-01-07
Disable Buildkite for dmd-cxx

wilzbach:
FYI @ibuclaw

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+# Don't run Buildkite for the dmd-cxx branch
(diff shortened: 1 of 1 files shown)


ibuclaw:
Missing an early return?

wilzbach:
> Missing an early return?

Argh. Yes. Thanks! It was definitely too late when I opened up this PR.

ibuclaw:
```
buildkite.sh: line 6: return: can only `return' from a function or sourced script
```
;-)

wilzbach:
Argh -> https://github.com/dlang/ci/pull/374

PR #377, opened 2019-01-13
Revert "Disable Lubeck as it fails with dependency errors"

wilzbach:
Reverts dlang/ci#376

So that it doesn't get lost, but lubeck has a bad track history now and I don't feel too inclined to have a project here which releases things with a broken CI on the project :/


ZombineDev:
I think we can now reenable it.

PR #380, opened 2019-01-23
run dlang-bot tests earlier

Basile-z:
It's the last project to finish these days, maybe at some point its tests became slower than they used to be.


buildkite.sh
```diff
@@ -125,6 +125,7 @@ projects=(
     "vibe-d/vibe-core+select" # 3m30s
     "higgsjs/Higgs" # 3m10s
     "rejectedsoftware/ddox" # 2m42s
+    "dlang-bots/dlang-bot" # 2m39s
```
ZombineDev: It just finished at 4m 54s - I'll move it up.

PR #381, opened 2019-04-05
ocean: Do not run d2conv

Geod24:
Starting from v5.0.0, ocean dropped D1 support and is D2 only.

CC @mihails-strasuns @leandro-lucarella-sociomantic


ZombineDev:
Thanks, guys!

PR #390, opened 2019-06-04
Add sociomantic-tsunami/swarm project

matthias-wende-sociomantic:
This is a naive attempt to add swarm to the dlang ci, now that it is a D2 project .

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -121,6 +121,7 @@ projects=(
     "vibe-d/vibe.d+libasync-base" # 3m45s
     "dlang/phobos" # 4m50s
     "sociomantic-tsunami/ocean" # 4m49s
+    "sociomantic-tsunami/swarm"
     "sociomantic-tsunami/turtle"
     "dlang/dub" # 3m55s
     "vibe-d/vibe-core+epoll" # 3m38s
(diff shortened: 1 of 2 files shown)


Geod24:
Side note: You might want to take care of [those deprecations](https://buildkite.com/dlang/ci/builds/659#ad1c5faa-e213-4f1e-8943-046a677fbc6b/134-169) before they turn into errors :)

buildkite.sh
```diff
@@ -121,6 +121,7 @@ projects=(
     "vibe-d/vibe.d+libasync-base" # 3m45s
     "dlang/phobos" # 4m50s
     "sociomantic-tsunami/ocean" # 4m49s
+    "sociomantic-tsunami/swarm"
```
leandro-lucarella-sociomantic: I don't know how important is to keep the time sorting....

buildkite/build_project.sh
```diff
@@ -211,7 +211,9 @@ case "$REPO_FULL_NAME" in
         dub test "--compiler=$DC"
         ;;
 
-    sociomantic-tsunami/ocean|sociomantic-tsunami/turtle)
+    sociomantic-tsunami/ocean | \
+    sociomantic-tsunami/turtle | \
+    sociomantic-tsunami/swarm)
```
leandro-lucarella-sociomantic: This could have been done more future proof by using a case like `sociomantic-tsunami/*)`.

matthias-wende-sociomantic: I might do that in the next PR, when adding the dhtproto.

PR #392, opened 2019-06-11
deactivate GFM

Basile-z:
[This commit](https://github.com/d-gamedev-team/gfm/commit/bd0dcd09497a2c1f421b2aebef40430ca6bdeeea#commitcomment-33877820) broke their own CI I believe. ping @d-gamedev-team, if you can fix from your side instead.


wilzbach:
Ping @p0nce

Basile-z:
Is it supposed to be effective already ?

PR #395, opened 2019-07-03
Revert "Disable Mecca pending PR merge"

wilzbach:
Reverts dlang/ci#393

CC @Geod24

PR #397, opened 2019-07-15
Disable Mecca once again

Geod24:
Reverts dlang/ci#395

Unfortunately the fix is not enough, apparently... And terrible to test.
Personally I'm fed up with fixing linking errors triggered by access to private symbols.


wilzbach:
Yeah we shouldn't cater for this at all.

PR #400, opened 2019-08-03
Revert "Temporarily remove optional"

wilzbach:
Reverts dlang/ci#399


ZombineDev:
@wilzbach Is this good to go now?

wilzbach:
Not sure if the dub registry got fixed or this is just a PR where we were lucky...

PR #401, opened 2019-08-10
Remove Basile-z/iz from the list of packages

Geod24:
The user apparently deleted their Github account.

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -150,7 +150,6 @@ projects=(
     "dlang-community/D-YAML" # 15s
     "libmir/mir-random" # 13s
     "dlang-community/libdparse" # 13s
-    "Basile-z/iz" # 12s
     "aliak00/optional" # 12s
     "dlang-community/dfmt" # 11s
     "jacob-carlborg/dlp" # 11s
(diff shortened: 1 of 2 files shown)


Geod24:
Also disabling `aliak/optional` as it seems packages can't be fetched from Gitlab.
CC @aliak00

aliak00:
@Geod24 Gitlab? You mean github? -> https://github.com/aliak00/optional
Why can't it be fetched?

wilzbach:
One of your dependencies is on gitlab and the dub registry seems to crash on gitlab dependencies.

aliak00:
Oh. That'd be silly then. Is it a dub problem? Because locally at least (i just deleted the package from ~/.dub) running dub test fetched from gitlab without problems?

wilzbach:
> Oh. That'd be silly then. Is it a dub problem?

Yeah it's really silly and a dub registry problem, but we had to revert it a week ago already (https://github.com/dlang/ci/pull/399).
Here's an example: https://buildkite.com/dlang/dmd/builds/8798#c7f66aa9-d746-4f7e-8d3f-d2d981088e4f

The problem is that if you go to https://code.dlang.org/packages/silly/0.8.2.zip, you directly get a 502 from the dub registry :/
I opened an issue here, but I guess a workaround for now would be to register a GitHub fork (`silly-gh`) :/

aliak00:
Boh. Ok. Thanks for the info. I'll ping here if I get around to forking
silly on gh as a workaround or something.

wilzbach:
> I'll ping here if I get around to forking silly on gh as a workaround

I think @FeepingCreature has expressed interest in implementing `git` support for `dub`, then you could just use sth. like `git://gitlab.com/../silly.git` and it would never even query the dub registry.

PR #404, opened 2019-12-18
reduce excessive retries

MartinNowak:
- use only a single retry to workaround excessive time spent on actual failures


Geod24:
Would there be a way to have those retry more spaced in time?

PR #405, opened 2020-01-06
Dub: Remove workaround, use Travis' script

Geod24:
Workaround unnecessary since https://github.com/dlang/dub/pull/1854
The exact call to `travis-ci.sh` blocks https://github.com/dlang/dub/pull/1853

PR #406, opened 2020-01-07
Fix dub testing on stable

Geod24:
Since the script is not versioned, those kinds of errors are quite common.
Using `git merge-base` seems like the most resilient option, as it won't break in the future.


thewilsonator:
https://github.com/dlang/ci/pull/405#issuecomment-571017275

Geod24:
So much for that. `git clone -b master` results in a shallow commit (only 1 commit in the history), hence the `git merge-base` fails.
Although it seems that the clone is unnecessary but I'll keep that for a later PR.

PR #409, opened 2020-01-25
Add a Dockerfile to build an agent

Geod24:
```
This Dockerfile will allow administrator to quickly and easily deploy an agent.
The dependencies were extracted from ansible/roles/test_deps/tasks/main.yml.
A user can run the agent with:
$ docker build -t buildkite_agent buildkite
$ docker run -e BUILDKITE_AGENT_TOKEN='AGENT-TOKEN-HERE' buildkite_agent
Note that since this image is not intended to be distributed, the usual cleanup isn't done
(and no attempt was made to minimize the dependencies).
```

CC @wilzbach @PetarKirov : I'm running this on my server now so if any issue shows up we'll know soon enough. Might want to give it a few days to see if nothing fails (@wilzbach originally spotted failures because Ocean has a hard dependency on OpenSSL 1.0.0 and Arch only provides 1.1).

### DIFF
--- a/ansible/roles/test_deps/tasks/main.yml
+++ b/ansible/roles/test_deps/tasks/main.yml
@@ -47,7 +47,6 @@
   with_items:
     - https://bintray.com/sociomantic-tsunami/dlang/download_file?file_path=libebtree6_6.0.s7-rc5-xenial_amd64.deb
     - https://bintray.com/sociomantic-tsunami/dlang/download_file?file_path=libebtree6-dev_6.0.s7-rc5-xenial_amd64.deb
-    - https://bintray.com/sociomantic-tsunami/dlang/download_file?file_path=d1to2fix_0.10.0-alpha1-xenial_amd64.deb
 - name: disable mongodb journaling
   lineinfile: { dest: /etc/mongodb.conf, regexp: "^{{ item.key }} *=", line: "{{ item.key }} = {{ item.value }}" }
   with_dict:
--- a/buildkite/Dockerfile
+++ b/buildkite/Dockerfile
@@ -1,0 +1,13 @@
+FROM buildkite/agent:ubuntu
+
+ADD travis_get_script /usr/local/bin/get_travis_test_script
+RUN apt-get update && apt-get install -y build-essential cmake curl gdb git jq libblas-dev \
+    libbz2-dev libcairo-dev libcurl4-gnutls-dev libevent-dev libgcrypt20-dev libgpg-error-dev \
+    libgtk-3-0 liblapack-dev liblzo2-dev libopenblas-dev libreadline-dev libssl-dev libxml2-dev \
+    libxslt1-dev libzmq3-dev mongodb-server moreutils ninja-build \
+    pkg-config python-dev python-yaml python3-nose redis-server \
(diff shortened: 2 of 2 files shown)


Geod24:
So turns out Travis always use 16.04 which has the old (1.0) openssl, but the upstream buildkite image uses the new one (1.1).
Sociomantic has a hard dependency on 1.0 so we can't use Bionic. On the other hand, if someone was to have a hard dependency on 1.1, it'd be game over. We really could use a way for project to install dependencies in a container.

wilzbach:
https://buildkite.com/dlang/ci/builds/706#5b2589d0-7fb7-450d-9dc7-b5f4c59ba112

The dlang-tour/core failure is because it's latest tag is using vibe.d 0.8.4. I will tag a new one after https://github.com/dlang-tour/core/pull/745/ is merged and then vibe.d should auto-detect the openssl version.
Similar story for https://github.com/rejectedsoftware/ddox/pull/221.

> Sociomantic has a hard dependency on 1.0 so we can't use Bionic. 

Buildkite allows for agent selection/filtering, so we could in theory make sure that Sociomantic libs are only tested on OpenSSL 1.0 agents.

> We really could use a way for project to install dependencies in a container.

Yeah, this project grew a bit that it's original "simply run dub test" idea. IIRC the original idea was to
- replicate the project's TravisCI config + setup as that's what most projects use, so the chances of it breaking are lower
- avoid dependency install to make builds faster + reduce traffic/bandwidth costs

Anyhow, I guess if we would allow projects to install their own dependencies we (1) need all agents to run as Docker agents and (2) parse their Travis config or similar :/

wilzbach:
> The dlang-tour/core failure is because it's latest tag is using vibe.d 0.8.4. I will tag a new one after dlang-tour/core#745 is merged and then vibe.d should auto-detect the openssl version.
Similar story for rejectedsoftware/ddox#221.

FYI: Both PRs are merged and released now.

Geod24:
Note: I fixed a few things on my runner, and it's currently the only runner in Buildkite.
Don't know what happened to the others. My server has 99.99% uptime so it shouldn't be a problem (except for potential disk space issues).

I definitely think we need something more robust for Buildkite. Perhaps we could explore Github actions ? We currently have a workflow setup in our project: https://github.com/bpfkorea/agora/blob/acbd5464cd735d6c62910d49e2bac7346926ceb0/.github/workflows/github_push.yaml
For DMD to test our project, they would just need to use this and replace the step where DMD is installed (a YAML parser could do that).
Then using https://github.com/actions/runner to run the job.

MartinNowak:
Yes, Github Actions work fairly well, for my uses.

You could do

build job -> [matrix](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) project job (with [needs](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) dependency on the build job)

Not sure if that would help anything with the proclaimed robustness issues in Buildkite.
I think most moving parts come from the projects themselves.

Github Actions come with (free) existing runners. If that promise holds in the long-run, then you should avoid the hassle to run your own.

PR #411, opened 2020-02-24
add Martin Nowak's std.io project

schveiguy:
Just got the project building with the latest dmd. I want to add it here so it doesn't get broken again.

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -184,6 +184,7 @@ projects=(
     "repeatedly/mustache-d"
     "s-ludwig/std_data_json"
     "s-ludwig/taggedalgebraic"
+    "MartinNowak/io"
 )
 # Add all projects that require more than 3GB of memory to build
 declare -A memory_req
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -306,6 +306,13 @@ case "$REPO_FULL_NAME" in
         use_travis_test_script
         ;;
 
+    MartinNowak/io)
+        export BUILD_TOOL=dub
+        export COVERAGE=false
+        export SKIP_IPv6_LOOPBACK_TESTS=yes
+        use_travis_test_script
+        ;;
+
     *)
         use_travis_test_script
         ;;


schveiguy:
How do I fix the build error? Travis CI works fine.

schveiguy:
So it looks like because I have meson and dub builds (recently added), the build script is having errors, I'm assuming because it's called without any parameters.

Is there a way to add parameters/env to the call? I read that I can include a specific buildkite.sh script. Do I need to do anything extra besides call travis.sh the right way?

wilzbach:
> Is there a way to add parameters/env to the call?

Scroll down within the same file for special cases. By default it tries to parse the Travis yaml and execute that one.

schveiguy:
@wilzbach, my script-fu is poor, and I don't really understand what is happening here. It appears that it's calling travis.sh without giving it a directive to use dub or meson. See the failure line here:

```
./travis.sh: line 11: BUILD_TOOL: unbound variable
```

That line is here: https://github.com/MartinNowak/io/blob/master/travis.sh#L11

I think the yml file is supposed to be adding that BUILD_TOOL variable. I don't know what buildkite is doing, or if it's just ignoring that directive. How do I get it to run the right thing?

wilzbach:
Ah I'm sorry. I didn't realize the files were moved recently.
You want to define a special case for the project here and e.g. do something like:

```
MartinNowak/io)
  export BUILD_TOOL=dub
  use_travis_test_script
  ;;
```

https://github.com/dlang/ci/blob/master/buildkite/build_project.sh

I hope that helps.

wilzbach:
Though the easiest fix would be do just add a default fallback to the Travis build script:

https://github.com/MartinNowak/io/pull/36

schveiguy:
OK, hopefully this works.

schveiguy:
Don't merge yet, I stuck the io project in first so I don't have to wait for the whole thing to find any more stupid problems. If it works, I'll reorder it back to the end.

schveiguy:
ugh didn't work. I still have to wait. Will just move it back.

schveiguy:
Weird, it now moved it to the top, even though I've moved it back down. But it now passes, good to merge @wilzbach

PR #414, opened 2020-03-19
Vibe.d: Clean up and remove non-Vibe-core backend

Geod24:
Those backends were deprecated, and are gone in master, so we're essentially building the same backend multiple times. Also removed a few outdated comment and re-activated a test that is fixed.

CC @s-ludwig
BTW there are a few `vibe-core` tests in vibe-d, should they be moved over to the new repository ?

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -109,16 +109,11 @@ esac
 ################################################################################
 projects=(
     # sorted by test time fast to slow (to minimize pending queue length)
-    "vibe-d/vibe.d+libevent-examples" # 12m1s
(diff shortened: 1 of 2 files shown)


s-ludwig:
@Geod24: Good point regarding the tests, I missed those. I'll remove all duplicates and add those that are only in the vibe.d repository to vibe-core. The question is then, should vibe-core be tested on Buildkite, too? With regards to resource/time use I'd think that just vibe.d should still be enough to guarantee a reasonable coverage.

Geod24:
Should be green now

PR #420, opened 2020-07-14
BuildKite: Don't use LDC *master* (but the latest tag instead)

kinke:
As a band-aid for the CI regression (due to LDC master now requiring LLVM 6+). Note that the next beta is scheduled for this evening, so someone should look into raising the installed LLVM version ASAP.

### DIFF
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -233,9 +233,7 @@ case "$REPO_FULL_NAME" in
         ;;
 
     ldc-developers/ldc)
-        cd runtime && git clone --depth 5 https://github.com/ldc-developers/druntime
-        git clone --depth 5 https://github.com/ldc-developers/phobos
-        cd .. && git submodule update
+        git submodule update --init
         mkdir bootstrap && cd bootstrap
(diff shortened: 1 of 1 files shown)


dnadlinger:
Also see https://github.com/dlang/ci/pull/421.

Geod24:
> ldc/intrinsics.di(54): Error: undefined identifier `LLVM_version`
> --
>  | ldc/intrinsics.di(54): Error: pragma `LDC_intrinsic` requires exactly 1 string literal parameter

Looks like a submodules / ldc mismatch ?

Geod24:
I added a commit to just use `git submodule update --init` instead of clone, let's see if that works...

Geod24:
Thanks!

kinke:
Thank *you* for the submodule fix.

WalterBright:
Thanks everyone for clearing up this mysterious blocker. I presume this was what was causing me problems with https://github.com/dlang/druntime/pull/3159 and https://github.com/dlang/dmd/pull/11413

WalterBright:
Yep, it's cleared up now!

PR #421, opened 2020-07-14
ansible: Bump LLVM to 8.0

dnadlinger:
LLVM 6.0+ will be required from the next LDC release
(and is already required for LDC master).

### DIFF
--- a/ansible/roles/test_deps/tasks/main.yml
+++ b/ansible/roles/test_deps/tasks/main.yml
@@ -31,7 +31,7 @@
     - libxml2-dev
     - libxslt1-dev
     - libzmq3-dev
-    - llvm-3.9-dev # ldc
+    - llvm-8-dev # ldc
     - mongodb-server
     - moreutils # sponge et.al.
     - ninja-build # ldc


Geod24:
I think dstep use LLVM too, so it might affect it (CC @jacob-carlborg ).
IIUC, this needs some action by @MartinNowak too.

dnadlinger:
I don't have the setup to easily test this locally right now, unfortunately.

jacob-carlborg:
> I think dstep use LLVM too, so it might affect it 

Yes, DStep uses LLVM, or specifically, libclang. It already supports 8.0.0. But anything newer has not been tested.

dnadlinger:
Seems like `libclang-8-dev # dstep` was already used.

kinke:
> Found LLVM: /usr/lib/llvm-3.9 (found suitable version "3.9.1", minimum required is "3.9")

Seems like it doesn't find the new installation. The LLVM dir/version isn't specified explicitly in the CMake cmdline (https://github.com/dlang/ci/pull/420/files#diff-e6bd4021b60e04e45022336bda968fdeL240-L244), so it should default to using the latest installed version it can find.

dnadlinger:
I don't think the changes would take effect automatically in PR testing.

kinke:
Yeah seems like it. So the next LDC beta has to be delayed until this is confirmed to work and deployed, unless another non-latest LDC tag is used in the meantime. Geez, I hate to depend on others' CI setup.

jacob-carlborg:
@kinke just disable LDC here, if you don't want to get delayed.

kinke:
Looking at https://github.com/dlang/ci/blob/b823197a15569f4ebeb5c107cdae55e2f1085508/buildkite/build_project.sh#L26, beta tags should be filtered out, so most likely no problem until the final is released, which will take a while. We'll find out soon enough, as I'm about to tag.

dnadlinger:
So how do we actually test this? Does anybody have a local test setup for this?

Geod24:
Waiting to hear back from @MartinNowak which I suspect is the only one able to deploy this (so even merging wouldn't get us far). @wilzbach can you confirm perhaps ?

MartinNowak:
Sebastian should also have access to the Hetzner Cloud project. Just let me know if someone else wants access.

Geod24:
I wouldn't mind.

kinke:
> hope this didn't break the current tests :)

It did. ;) - #422 should be pulled ASAP.

PR #423, opened 2020-08-09
Add autodecode test to buildkite.

schveiguy:
This is the next step in the process. In order to test autodecode PRs actually work and don't break later, we need to be able to test modules with no autodecode enabled as they are fixed.

See plan in https://github.com/dlang/phobos/pull/7586

### DIFF
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -264,6 +264,12 @@ case "$REPO_FULL_NAME" in
             make -C druntime -j2 -f posix.mak
         fi
         cd "$(basename "${REPO_FULL_NAME}")"&& make -f posix.mak clean && make -f posix.mak -j2 buildkite-test
+        if [ "$REPO_FULL_NAME" == "dlang/phobos" ] ; then
+            # test autodecode items
+            export NO_AUTODECODE=1
+            echo "--- Running no-autodecode test"
+            make -f posix.mak clean && make -f posix.mak -j2 autodecode-test
+        fi
         rm -rf "$TMP"
         ;;
 
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -44,6 +44,7 @@ case "$REPO_FULL_NAME" in
     "dlang/dmd" | \
     "dlang/druntime" | \
     "dlang/phobos" | \
+    "dlang/phobos+no-autodecode" | \
     "dlang/tools" | \
     "dlang/dub" | \
     "dlang/ci")
@@ -58,7 +59,8 @@ case "$REPO_FULL_NAME" in
             case "$REPO_FULL_NAME" in
                 "dlang/dmd" | \
                 "dlang/druntime" | \
-                "dlang/phobos")
+                "dlang/phobos" | \
+                "dlang/phobos+no-autodecode")
                 ref_to_use="IS-ALREADY-CHECKED-OUT"
                 ;;
             *)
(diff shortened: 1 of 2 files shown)


MoonlightSentinel:
Why build phobos twice? It's already tested with autodecoding on the other CIs.

schveiguy:
What?!!! where?

schveiguy:
And this isn't a full unittest build, it only tests the specified modules.

buildkite/build_project.sh
```diff
@@ -264,6 +264,11 @@ case "$REPO_FULL_NAME" in
             make -C druntime -j2 -f posix.mak
         fi
         cd "$(basename "${REPO_FULL_NAME}")"&& make -f posix.mak clean && make -f posix.mak -j2 buildkite-test
```
MoonlightSentinel, line 266: > What?!!! where?

This builds phobos as a dependency of `buildkite-test`

schveiguy: I need Phobos to be built with this set to false:
https://github.com/dlang/phobos/blob/ba02646078ba45c9ad26ea348d0e9d7fad8f6a68/std/range/primitives.d#L2412

If there's already a CI that does this, I'll use it instead.

schveiguy: Note, the `-version=NoAutodecodeStrings` directive does nothing for current Phobos. That is what I'm going to use to conditionally set that bool to false. But in order for things to test properly, I need a CI set up that will do it.

MoonlightSentinel: I know, my  initial thought was to only build phobos with disabled autodecoding. But that doesn't work because the `betterC` tests were migrated from CircleCI to BuildKite ... guess we'll have to accept the double penalty.

schveiguy:
> Why build phobos twice? It's already tested with autodecoding on the other CIs.

I misunderstood this. You mean build it once with autodecode disabled, and let other CIs test the normal situation. I had thought you meant that it's already testing with autodecode removed on other CIs.

Yeah, it would be nice to just do one test here, or somewhere else. I'm going the cheap and easy route, and once it's working, we can look at moving things around.

schveiguy:
OK, this should be good to merge. I used echo to separate out the no-autodecode output for the test (which does nothing, but of course will once I start popuplating modules in there).

wilzbach:
You might want to consider adding it as a separate project configuration. We do the same with Vibe-d, e.g. `dlang/phobos+no-auto-decode`. This has the advantage of being able to run in parallel and a better error message as people then know why Phobos failed on Buildkite.

schveiguy:
Yes thank you @wilzbach! I will follow that model.

schveiguy:
OK, I've changed to test in its own project, and it should be good now.

schveiguy:
Still need this merged to go further on autodecode.

buildkite/build_project.sh
```diff
@@ -267,6 +269,23 @@ case "$REPO_FULL_NAME" in
         rm -rf "$TMP"
```
wilzbach, line 269: Why not add dlang/phobos+no-autodecode to the list above and then distinguish at the end?
(everything else just looks duplicated)


```suggestion
        cd "$(basename "${REPO_FULL_NAME}")"&& make -f posix.mak clean 
        if [ "$REPO_FULL_NAME" == "dlang/phobos+no-autodecode" ] ; then
          make -f posix.mak -j2 autodecode-test
        else
          make -f posix.mak -j2 buildkite-test
        fi
        rm -rf "$TMP"
```

schveiguy: I need to build phobos with `NO_AUTODECODE=1`. It's looking more and more like a mess if I have another `if` above.

schveiguy: If this is what it takes to get it merged, I can do it. Just let me know.

schveiguy:
Thanks!

buildkite/build_project.sh
wilzbach: I have no strong preference/objection.

PR #424, opened 2020-08-14
Temporarily disable dlang-community/libdparse

Geod24:
Release v0.16.0 requires an xml lib which broke the build.
See dlang-community/libdparse#431.

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -144,7 +144,7 @@ projects=(
     "libmir/mir-algorithm" # 17s
     "dlang-community/D-YAML" # 15s
     "libmir/mir-random" # 13s
-    "dlang-community/libdparse" # 13s
+    #"dlang-community/libdparse" # 13s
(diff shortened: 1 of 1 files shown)


MoonlightSentinel:
For the future, could we enhance the scripts to allow explicit tags s.t. we can freeze on an earlier version instead of outright disabling a project?

wilzbach:
> For the future, could we enhance the scripts to allow explicit tags s.t. we can freeze on an earlier version instead of outright disabling a project?

https://github.com/dlang/ci/blob/d7ffcd3823883aac6869b0ac784a4cd23483827f/buildkite/build_project.sh#L31-L34

MoonlightSentinel:
Nevermind, maybe freeze `libdparse` on `0.15.4` then?

wilzbach:
Closed in favor of https://github.com/dlang/ci/pull/425

PR #427, opened 2020-08-16
Use newer bootstrap.sh to build DMD

wilzbach:
Can be merged after https://github.com/dlang/dmd/pull/11580 is in stable and master.

Not necessary as backwards compatibility should be provided (for now), but gets rid of deprecation messages and allows to remove the posix.mak file one day.


Geod24:
Rebase ?

PR #428, opened 2020-08-25
Re-enable unit-threaded

atilaneves:


MoonlightSentinel (approved):

Geod24:
```
~> bash
--
  | TERM environment variable not set.
```

atilaneves:
I think the failure was caused by the CI scripts clearing the console but TERM isn't set when run from buildkite. I just pushed a new version of unit-threaded so here's hoping this passes now, especially since it being disabled allowed a regression to sneak through to dmd.

Geod24:
@atilaneves : You need to tag a new release, Buildkite doesn't use master.

atilaneves:
Debugging why this fails is "fun". Googled and tried but can't reproduce a local environment with TERM not set. Except maybe a cron job?

MoonlightSentinel:
Commented on the commit, your workaround does not work with `set -u`.

atilaneves:
> Commented on the commit, your workaround does not work with `set -u` AFAICT

Any ideas?

MoonlightSentinel:
https://github.com/atilaneves/unit-threaded/commit/4213bfb9b5b06f541281e692688e81fb9090ec65#r41781457

Geod24:
I see you gave up :D

PR #429, opened 2020-08-26
Use a more recent ocean version when testing swarm

MoonlightSentinel:
Some required bugfixes (e.g sociomantic-tsunami/ocean#831) were already
merged into ocean but are not yet included into an official release.

This unblocks dlang/dmd#8429 for now and should be reverted when
a new release is tagged.

### DIFF
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -33,6 +33,16 @@ case "$REPO_URL" in
         # Use master as Vibe.d covers a lot of the language features
         ref_to_use=master
         ;;
+
+    https://github.com/sociomantic-tsunami/swarm)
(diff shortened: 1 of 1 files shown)


wilzbach:
Shouldn't turtle be included as well?

MoonlightSentinel:
Right, added turtle as well. Let's see if that works.

wilzbach (approved):

MoonlightSentinel:
Both passed!

MoonlightSentinel:
CC @Geod24

MoonlightSentinel:
Swarn & Turtle got new releases so this isn't needed anymore.

PR #432, opened 2020-09-23
Add autowrap to the list of projects

atilaneves:
[Autowrap](https://github.com/symmetryinvestments/autowrap) is an important project for Symmetry, and not only that, apparently a good canary for template instantiation issues that manifest as linker errors due to missing symbols.

The removal of the `-debug` hack from dmd broke its test suite, but [this commit](https://github.com/dlang/dmd/commit/5323f86d0ddb6f1c8384ed5500a1b32134bdcd02) fixed it. I want to add it to buildkite so that we catch any future regressions.

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -109,6 +109,7 @@ esac
 ################################################################################
 projects=(
     # sorted by test time fast to slow (to minimize pending queue length)
+    "symmetryinvestments/autowrap" #>10min
     "vibe-d/vibe.d+examples" # 9m51s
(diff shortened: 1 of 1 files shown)


MoonlightSentinel (approved):

Geod24:
The Travis script seems quite advanced... Some of those packages aren't available, and the env variables aren't set.

alexandrumc:
I think this can be closed now: https://github.com/dlang/ci/commit/05df6423c2a78a4e410030b793e1e31e7ab1ed73

PR #442, opened 2021-02-26
Fix ocean following v6.0.0 release

Geod24:


Geod24:
Setting automerge myself since Buildkite is currently broken.

PR #450, opened 2021-12-09
Reduce a bit of the soul-crushing log spam created by buildkite

andralex:
First, the default git clone will output progress message every percent of the way, which is not useful for offline log inspection.

Second, buildkite uses the `---` prefix to group log output. However, that's not nested so calling the cloning script will group things the wrong way. So I added those `echo`s to distinguish cloning from building.

Related: isn't it excessive to output every single command run? I'd understand doing so temporarily for debugging (and even in that case I'd argue well-thought logging would be vastly better), but to leave it on by default seems much.


maxhaton (approved):

andralex:
@maxhaton thanks! Is that script deployed automatically?

Geod24:
Yes it is.

PR #457, opened 2022-07-09
Allow to build both pre-merge and post-merge druntime

Geod24:
We need to keep pre-merge logic around for build_distribution,
which is not too much logic.

### DIFF
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -57,6 +57,7 @@ echo "--- Checking ${REPO_FULL_NAME} for a core repository and branch merging wi
 # Don't checkout a tagged version of the core repositories like Phobos
 case "$REPO_FULL_NAME" in
     "dlang/dmd" | \
+    "dlang/druntime" | \
     "dlang/phobos" | \
     "dlang/phobos+no-autodecode" | \
     "dlang/tools" | \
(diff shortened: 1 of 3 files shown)


buildkite/build_project.sh
```diff
@@ -295,11 +298,19 @@ case "$REPO_FULL_NAME" in
         export TMPDIR="$TMP"
         rm -rf "$TMP" && mkdir -p "$TMP"
         # patch makefile which requires gdb 8 - see https://github.com/dlang/ci/pull/301
-        sed "s/TESTS+=rt_trap_exceptions_drt_gdb//" -i dmd/druntime/test/exceptions/Makefile
+        if [ -f dmd/druntime/test/exceptions/Makefile ]; then
+            sed "s/TESTS+=rt_trap_exceptions_drt_gdb//" -i dmd/druntime/test/exceptions/Makefile
+        elif [ -f druntime/test/exceptions/Makefile ]; then
+            sed "s/TESTS+=rt_trap_exceptions_drt_gdb//" -i druntime/test/exceptions/Makefile
+        fi
```
ibuclaw, line 305: I can't say why you needed to revert it all, but on a second look, there's a thinko here, where I should have kept in the if condition.

Geod24: Because I saw the CI checking out `master` and building it.
I figured best way forward was to:
1) Keep support for building the old scheme
2) Add support for the new scheme
3) Remove `druntime` as a project we test

PR #463, opened 2022-08-27
Disable intel-intrinsics

Geod24:
Latest release broke the CI.
Issue here: https://github.com/AuburnSounds/intel-intrinsics/issues/99

CC @p0nce


p0nce:
CC @WalterBright This is a DMD regression (worked in 2.097) and now build Kite catch the regression.

PR #467, opened 2022-10-03
Update downloads.dlang.org links to use https

ibuclaw:
This is now possible now the site now has a new front for its bucket.

### DIFF
--- a/buildkite/build_release.sh
+++ b/buildkite/build_release.sh
@@ -19,7 +19,7 @@ git fetch upstream $("./$DIR/origin_target_branch.sh" https://github.com/dlang/i
 git checkout FETCH_HEAD --quiet
 
 mkdir build
-LATEST=$(wget -q -O - http://downloads.dlang.org/releases/LATEST)
+LATEST=$(wget -q -O - https://downloads.dlang.org/releases/LATEST)
 rdmd build_all.d v$LATEST $BRANCH
 rm -f build/*.zip # only keep lzma archives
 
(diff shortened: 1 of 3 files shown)


WalterBright (approved):

buildkite/build_release.sh
```diff
@@ -34,9 +34,9 @@ aws --profile ddo s3 sync dmd-$BRANCH-$TODAY/ s3://downloads.dlang.org/nightlies
 
 # create redirects from dmd-$BRANCH/* to dmd-$BRANCH-2018-10-12/*
 for file in $(ls dmd-$BRANCH-$TODAY); do
-    echo -n | aws --profile ddo s3 cp - s3://downloads.dlang.org/nightlies/dmd-$BRANCH/$file --acl public-read --cache-control max-age=604800 --website-redirect http://downloads.dlang.org/nightlies/dmd-$BRANCH-$TODAY/$file
+    echo -n | aws --profile ddo s3 cp - s3://downloads.dlang.org/nightlies/dmd-$BRANCH/$file --acl public-read --cache-control max-age=604800 --website-redirect https://downloads.dlang.org/nightlies/dmd-$BRANCH-$TODAY/$file
```
Geod24, line 37: Hum we're not pointing to anymore ?

Geod24:
I think we should just nuke `build-release` ?

buildkite/build_release.sh
ibuclaw: Hmm, yes. That's right. The downloads.dlang.org record is no longer usable as an S3 bucket.

PR #474, opened 2023-02-09
Add sdc:sdfmt to buildkite projects.

maxhaton:


maxhaton:
Currently only the build is tested.

PR #477, opened 2023-06-02
Add symmetryinvestments' ldapauth project

alexandrumc:
Two of the projects that we regularly use internally.


Geod24:
Looks like this might need a few adaptations on the CI script

alexandrumc:
> Looks like this might need a few adaptations on the CI script

So it seems, though I have run them locally (followed https://github.com/dlang/ci/blob/master/README.md#how-to-build-a-project-locally) and it worked. Will look into it.

PR #482, opened 2023-06-07
Fix Higgs build by re-adding Python2

alexandrumc:
Higgs no longer compiles because we switched from python2 to python3 in Dockerfile. This is an attempt to fix it.

### DIFF
--- a/buildkite.sh
+++ b/buildkite.sh
@@ -233,11 +233,16 @@ cat << EOF
     ${DEFAULT_COMMAND_PROPS}
 EOF
 
-if [ "${memory_req["$project_name"]:-x}" != "x" ] ; then
+#if [ "${memory_req["$project_name"]:-x}" != "x" ] ; then
+#cat << EOF
+    #agents:
+      #- "memory=${memory_req["$project_name"]}"
+#EOF
+#fi
+
 cat << EOF
     agents:
-      - "memory=${memory_req["$project_name"]}"
+      - "testtag"
 EOF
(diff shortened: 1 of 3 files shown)


alexandrumc:
My first commit (where I targeted my local agent) shows that it actually works.

Geod24:
Not interested in fixing Higgs ? Supporting Python2 is only going to add more pain in the future.

alexandrumc:
> Not interested in fixing Higgs ? Supporting Python2 is only going to add more pain in the future.

The docs say that it requires Python2: https://github.com/higgsjs/Higgs/blob/master/README.md?plain=1#L35. Also, on your current setup `python` points to the Python3 binary, so one would have to explicitly use `python2` (as I did for Higgs) to actually use Python2.

Probably I could update this https://github.com/higgsjs/Higgs/blob/master/source/runtime/layout.py but I don't quite have time right now, will try to update it in the following days, so I would suggest moving on with this solution for now.

Geod24:
> The docs say that it requires Python2: https://github.com/higgsjs/Higgs/blob/master/README.md?plain=1#L35. Also, on your current setup `python` points to the Python3 binary, so one would have to explicitly use `python2` (as I did for Higgs) to actually use Python2.

That doc was written 9 years ago (https://github.com/higgsjs/Higgs/commit/017712c63d0e3ab3a395f8b6ce0e0a156d896147), 6 years before Python2 went EOL.

PR #495, opened 2025-09-08
buildkite/build_project.sh: Use --debug=ae_unittest with ae

### LINKED ISSUE #58: ae unit tests don't run on Buildkite
Currently, in D, when building a program and running tests, we don't really have a way to say "please build/run unit tests in this package but not other packages". 

This is a problem for libraries, because usually in a library, we only want to run the library's unit tests when we are developing the library, and not any time when the library is simply used as part of a bigger project.

We can sometimes work around this by building the library separately, and then linking to it. Then, the static or shared library being linked will never contain unit test code. However, this doesn't always work - e.g. it doesn't work for templates. However, there isn't really a universal solution.

To work around this issue, ae has all unit test blocks wrapped in a `debug (ae_unittest)` check, then runs `dub test -debug=ae_unittest` from its CI. This guarantees that unit tests are only compiled from ae's CI, and never in any programs that use ae.

However, this change had the unfortunate effect that ae's tests also no longer run when other parties do want to run ae's tests. Specifically, because ae is a non-trivial and long-lived collection of D code, and is used in some DFL projects (such as the forum, or the documentation CI system), it is part of the community project test suite (implemented using BuildKite), which tests a collection of community projects for every proposed change in D itself.

I believe this is the script used to test an arbitrary community project:
https://github.com/dlang/ci/blob/master/buildkite/travis_get_script

Because it runs `dub test` without `-debug=ae_unittest`, it is only able to test a much smaller surface area.

As a consequence of this, the degradation in testing has allowed regressions to be introduced in D that broke ae (and thus projects which use ae): https://github.com/dlang/dmd/issues/21826 . These regressions were only noticed long after the corresponding D release, when I tried to update the compiler to a version which includes these changes.

To resolve this, we need to either:

1. Figure out a better way to make `unittest` blocks only run when `ae` itself is being tested, and never when a project using `ae` is being tested. @Geod24 Could you offer any hints here, is there perhaps a Dub feature I've missed?
2. Improve compatibility with the BuildKite CI scripts so that they build ae with `-debug=ae_unittest`.

We need to be careful when making these changes to avoid causing disruption for the upstream D CI, even if doing so is simply because we are now exposing missed regressions.

CyberShadow:
Fixes https://github.com/CyberShadow/ae/issues/58

Expected to fail until the Druntime regression is addressed somehow.

### DIFF
--- a/buildkite/build_project.sh
+++ b/buildkite/build_project.sh
@@ -251,7 +251,7 @@ case "$REPO_FULL_NAME" in
     CyberShadow/ae)
         # remove failing extended attribute test
         perl -0777 -pi -e "s/unittest[^{]*{[^{}]*xAttrs[^{}]*}//" sys/file.d
-        use_travis_test_script
+        dub test "--compiler=$DC" --debug=ae_unittest
         ;;
 
     ikod/dlang-requests)


CyberShadow:
After many upstream fixes and workarounds in ae... the stars are finally in alignment again. Let's merge before another regression takes hold.

thewilsonator (approved):

CyberShadow:
Thanks!

thewilsonator:
No problems. Btw for non dmd/phobos repos, please do reach out earlier if I do not respond, I tend to lost track of PRs/issues in those a lot faster than the main repos.

